PR #31, opened 2020-05-15
feat: add EmailPasswordSignUpForm component

HellWolf93:
#### Changes proposed
- Add EmailPasswordSignUpForm component
- Add tests for EmailPasswordSignUpForm component
- Add docs for EmailPasswordSignUpForm component


cypress/integration/auth/emailPasswordSignUpForm.js
```diff
@@ -0,0 +1,37 @@
...
+            [],
+        ).as('firebaseSignUp');
+    });
+
+    it('should request create user endpoint', () => {
+        cy.get(NAME_INPUT).type('Test');
+        cy.get(EMAIL_INPUT).type('test@email.com');
+        cy.get(PASSWORD_INPUT).type('TestPassword');
+        cy.get(BUTTON).click();
+        cy.wait('@firebaseSignUp');
```
LeandroTorresSicilia, line 27: We should do assertions that check for something that indicates the user is authenticated like the new URL or an element that exists when authenticated.
Ideally, we should start up the emulator and login and then make the assertion but I don't know if right now there is an emulator for auth.

cypress/integration/auth/emailPasswordSignUpForm.js
```diff
@@ -0,0 +1,37 @@
...
+    });
+
+    it('should show spinner while loading', () => {
+        cy.get(NAME_INPUT).type('Test');
+        cy.get(EMAIL_INPUT).type('test@email.com');
+        cy.get(PASSWORD_INPUT).type('TestPassword');
+        cy.get(BUTTON).click();
+        cy.get(SPINNER_BACKDROP).should('exist');
```
LeandroTorresSicilia, line 35: In the case of unit test is a good practice make only one assertion for a test, but in the case of integrations we can make more than on assertion, then I think we can check for the spinner in the same test above

cypress/integration/auth/emailPasswordSignUpForm.js
```diff
@@ -0,0 +1,37 @@
...
+    });
+});
```
LeandroTorresSicilia, line 37: I think we could add another test case for when user type a wrong email or a too short password and then the error message is showed

packages/auth/docs/components/EmailPasswordSignUpForm.story.mdx
```diff
@@ -0,0 +1,32 @@
...
+<Meta title="Auth/Components" component={EmailPasswordSignUpForm} />
+
+# Overview
+
+`EmailPasswordSignUpForm` component is a fully functional Firebase 'email and password flow' sign up form.
+It will notify the errors occurred during the sign up process through app level messages.
+
+This component HAVE TO be used under a properly configured `RainbowFirebaseApp` component.
+
+### Translatable texts
```
LeandroTorresSicilia, line 14: Internationalization FormattedMessage keys

packages/auth/docs/stories/emailPasswordSignUpForm.story.js
```diff
@@ -0,0 +1,183 @@
...
+import WhenNoAuthenticated from '../../src/components/WhenNoAuthenticated';
+import EmailPasswordSignUpForm from '../../src/components/EmailPasswordSignUpForm';
+
+const Form = () => {
+    const style = {
+        margin: 'auto',
+        maxWidth: '400px',
+    };
+    const logoStyle = { textAlign: 'center' };
+    const logoTextStyle = { fontSize: '1.8em' };
```
LeandroTorresSicilia, line 15: since these styles are constants them can be defined outside the component

packages/auth/docs/stories/emailPasswordSignUpForm.story.js
```diff
@@ -0,0 +1,183 @@
...
+                <path
+                    fill="#F8752D"
+                    d="M2.032 13.995h7.84c1.105 0 2 .895 2 2v7.84c0 1.104-.895 2-2 2h-7.84c-1.105 0-2-.896-2-2v-7.84c0-1.105.895-2 2-2z"
+                />
+                <rect width="11.84" height="11.84" x="28.075" y="27.954" fill="#01B6F5" rx="2" />
+                <rect width="11.84" height="11.84" x="14.005" y="27.954" fill="#EA4243" rx="2" />
+                <rect width="11.84" height="11.84" x=".032" y="27.954" fill="#1AD1A3" rx="2" />
+            </g>
+        </svg>
+    );
```
LeandroTorresSicilia, line 41: same this can be outside or put this in another file to avoid pollute the example

packages/auth/docs/stories/emailPasswordSignUpForm.story.js
```diff
@@ -0,0 +1,183 @@
...
+        <div>
+            <div className="rainbow-p-around_large" style={logoStyle}>
+                {rainbowLogo}
+                <h3 style={logoTextStyle}>rainbow-modules</h3>
+            </div>
+            <Card className="rainbow-p-around_large" style={style}>
+                <EmailPasswordSignUpForm
+                    nameIcon={nameIcon}
+                    emailIcon={emailIcon}
+                    passwordIcon={passwordIcon}
```
LeandroTorresSicilia, line 159: since we don't define any props the icons should be part of the component for now

packages/auth/docs/stories/emailPasswordSignUpForm.story.js
```diff
@@ -0,0 +1,183 @@
...
+                <Form />
+            </WhenNoAuthenticated>
+            <WhenAuthenticated path="/app" redirect="/">
+                <span>Authenticated!</span>
+                <br />
+                <Button label="Log Out" onClick={() => app.auth().signOut()} />
+            </WhenAuthenticated>
+        </RainbowFirebaseApp>
+    );
+};
```
LeandroTorresSicilia, line 179: There is a way we can see all the example code in the documentation?

Screen Shot 2020-05-16 at 1 08 15 PM

packages/auth/src/components/EmailPasswordSignUpForm/index.js
```diff
@@ -0,0 +1,120 @@
...
+import { FormattedMessage, useIntl } from 'react-intl';
+import { showAppMessage, showAppSpinner, hideAppSpinner } from '../../../../app/src/actions';
+import useFirebaseApp from '../../../../firebase-hooks/src/useFirebaseApp';
+
+const EmailPasswordSignUpForm = (props) => {
+    const intl = useIntl();
+    const { className, style } = props;
+    const [name, setName] = useState('');
+    const [email, setEmail] = useState('');
+    const [password, setPassword] = useState('');
```
LeandroTorresSicilia, line 13: We should use redux-form which is setup in RainbowFirebaseApp like react-intl, then it should work out of the box.
Redux-form take care of the state of the form and also make the validation through the validate function.
Note that we should add test for the validate function and also add test cases for when errors show up.

packages/auth/src/components/EmailPasswordSignUpForm/__test__/emailPasswordSignUpForm.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    it('should show error message using app message', (done) => {
+        expect.assertions(1);
+        const component = mountWithIntl(<EmailPasswordSignUpForm />);
+        component.find('form').simulate('submit');
+        setTimeout(() => {
+            expect(actions.showAppMessage).toHaveBeenCalledTimes(1);
+            done();
+        }, 0);
```
LeandroTorresSicilia, line 48: can we test this with async await instead of using setTimeout?

packages/auth/src/components/EmailPasswordSignUpForm/index.js
```diff
@@ -0,0 +1,120 @@
...
+    const nameLabel = <FormattedMessage id="EmailPasswordSignUpForm.name" defaultMessage="Name" />;
+    const emailLabel = (
+        <FormattedMessage id="EmailPasswordSignUpForm.email" defaultMessage="Email Address" />
+    );
+    const passwordLabel = (
+        <FormattedMessage id="EmailPasswordSignUpForm.password" defaultMessage="Password" />
+    );
+    const buttonLabel = (
+        <FormattedMessage id="EmailPasswordSignUpForm.signup" defaultMessage="Create account" />
+    );
```
LeandroTorresSicilia, line 42: this const can be outside the component to avoid recreation of them in each rerender

packages/auth/src/components/EmailPasswordSignUpForm/index.js
```diff
@@ -0,0 +1,120 @@
...
+        defaultMessage: 'Enter your email',
+    });
+    const passwordPlaceholder = intl.formatMessage({
+        id: 'EmailPasswordSignUpForm.passwordPlaceholder',
+        defaultMessage: 'Enter your password',
+    });
+
+    const buttonStyle = {
+        width: '100%',
+    };
```
LeandroTorresSicilia, line 58: let's use styled-components for styles to maintain separation of concerns

packages/auth/src/components/EmailPasswordSignUpForm/index.js
```diff
@@ -0,0 +1,120 @@
...
+            <Input
+                label={nameLabel}
+                icon={nameIcon}
+                placeholder={namePlaceholder}
+                name="name"
+                value={name}
+                onChange={(event) => setName(event.target.value)}
+                required
+            />
+            <br />
```
LeandroTorresSicilia, line 71: it is better to use styles instead of creating dom element to separate components

packages/auth/src/helpers/intl-enzyme-test-helper.js
```diff
@@ -0,0 +1,23 @@
+import { IntlProvider } from 'react-intl';
```
LeandroTorresSicilia, line 1: Good job!!!
let's change the file name to `intlEnzyme`, we will be using camel case instead of dashes

packages/auth/src/components/EmailPasswordSignUpForm/index.js
```diff
@@ -0,0 +1,120 @@
...
+
+EmailPasswordSignUpForm.defaultProps = {
+    className: undefined,
+    style: undefined,
+    nameIcon: undefined,
+    emailIcon: undefined,
+    passwordIcon: undefined,
+};
+
+export default EmailPasswordSignUpForm;
```
LeandroTorresSicilia, line 120: we need to export this component in `packages/auth/src/index.js`

cypress/integration/auth/emailPasswordSignUpForm.js
HellWolf93: As for now, only Database, Firestore, Functions and Hosting are available for emulator. If we are going to test this it will be necesary to hit the real API, and handle the creation and deletion of users each time the tests runs, we could do that but I dont think it's not a good idea, given that we can trust firebase's ```createUserWithEmailAndPassword``` function will correctly authenticate the user when the sign up is correct

cypress/integration/auth/emailPasswordSignUpForm.js
```diff
@@ -0,0 +1,37 @@
+const FORM_URL = '/iframe.html?id=auth-stories--basic-email-password-sign-up-form';
+const BUTTON = 'button[type="submit"]';
+const SPINNER_BACKDROP = '[data-cy="app-spinner"]';
+const NAME_INPUT = 'input[name="name"]';
+const EMAIL_INPUT = 'input[name="email"]';
+const PASSWORD_INPUT = 'input[name="password"]';
+
+describe('EmailPasswordSignUpForm', () => {
+    beforeEach(() => {
```
LeandroTorresSicilia, line 9: can we do this in the before hook instead of in order to call it only once?

LeandroTorresSicilia:
Screen Shot 2020-05-16 at 12 08 42 PM
I think we should fix how icons look when the spinner shows up, also, since the icons are internals they should change its colors with the theme

HellWolf93:
> Screen Shot 2020-05-16 at 12 08 42 PM
> 
> I think we should fix how icons look when the spinner shows up, also, since the icons are internals they should change its colors with the theme

Fixed, just added `z-index` to the spinner container.

cypress/integration/auth/emailPasswordSignUpForm.js
HellWolf93: it is necessary to do this in beforeEach beause the server config is restored to its default between tests

packages/auth/src/components/EmailPasswordSignUpForm/index.js
```diff
@@ -0,0 +1,105 @@
...
+    const [name, setName] = useState('');
+    const [email, setEmail] = useState('');
+    const [password, setPassword] = useState('');
+    const app = useFirebaseApp();
+
+    const namePlaceholder = intl.formatMessage({
+        id: 'EmailPasswordSignUpForm.namePlaceholder',
+        defaultMessage: 'Enter your name',
+    });
+    const emailPlaceholder = intl.formatMessage({
```
reiniergs, line 32: can we move all these message components into a different file e.g. message.js then you can export then and that way the logic for I18n is isolated?

reiniergs: The <FormatedMessage /> components above could be moved out of this file also

HellWolf93: > can we move all these message components into a different file e.g. message.js then you can export then and that way the logic for I18n is isolated?

we should not do that because `intl` is obtained with useIntl, and hooks can only be called within components, and the placeholder should be a plain string

reiniergs: yeah, I hear you! ... I guess I was thinking about using `defineMessage` API. Why we need this is the question :) ... because moving forward I believe It will be nice to develop messages in a different file as a best practice, kinda separation of concern.

reiniergs: by the way of example:

```js
import {defineMessages, defineMessage} from 'react-intl';

export const messages = defineMessages({
  greeting: {
    id: 'app.home.greeting',
    description: 'Message to greet the user.',
    defaultMessage: 'Hello, {name}!',
  },
});
export const msg = defineMessage({
  id: 'single'
  defaultMessage: 'single message',
  description: 'header'
})
```

then you can use the hook and do the formatting in the component

more info here: https://formatjs.io/docs/react-intl/api

HellWolf93: thanks for the clarification and the example, i think it is OK now

Issue #30, opened 2020-05-13
feat: implement `SideBarNavigation` component


TahimiLeonBravo:
#### module
```@rainbow-modules/app```

##### Design
https://zpl.io/bAJWAA8
https://zpl.io/2EJWywM

#### Usage
```
<SideBarNavigation>
   <SideBarOption icon={} selectedIcon={} path="/app/import" label="" exact />
</SideBarNavigation>
```
#### Requirements
- [ ] it should navigate to the path when the user click the option
- [ ] it should select an option based on the browser url
- [ ] it should use icon as selectedIcon as fallback when the selectedIcon wasn't passed

#### Documentation
-  [ ] component API

#### Definition of Done
- [ ] component working as expected
- [ ] unit tests(jest)
- [ ] end to end tests(cypress.io)
- [ ] documented
    - stories
    - component overview(mdx under component folder)
            - what do you need to know if you are going to use this component?
    - prop table

PR #33, opened 2020-05-19
feat: add SideBarNavigation component

HellWolf93:
fix: #30


cypress/integration/app/sideBarNavigation.js
```diff
@@ -0,0 +1,22 @@
...
+
+    it('should navigate to the selected route', () => {
+        cy.get(NAVIGATION_ITEM).eq(1).click();
+        cy.location().should((loc) => {
+            expect(loc.pathname).to.equal('/export');
+        });
+    });
+
+    it('should select an option based on the browser url', () => {
+        cy.window().then((win) => {
```
reiniergs, line 17: can we use async/await syntax here?

packages/app/docs/stories/icons/rainbowLogo.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const RainbowLogo = (props) => {
```
reiniergs, line 4: This could be a candidate for an icon under @rainbow-modules/icons since it's going to be used everywhere, another PR off course but I will let @TahimiLeonBravo to make that call

packages/app/src/components/SideBarNavigation/index.js
```diff
@@ -0,0 +1,61 @@
...
+import { Sidebar } from 'react-rainbow-components';
+import { useLocation } from 'react-router-dom';
+import StyledContainer from './styled';
+
+const SideBarNavigation = (props) => {
+    const { className, style, children } = props;
+    const location = useLocation();
+    const routes = [];
+    const arrayChildren = React.Children.toArray(children);
+    arrayChildren.forEach((element, index) => {
```
reiniergs, line 12: let's use a `map` here instead of a `forEach`
```js
const routes= arrayChildren.map(() => {});
```

reiniergs: also, I don't think we need to implement that logic our self here, take a look at https://reacttraining.com/react-router/web/api/Hooks/useroutematch ... `match` API will solve the problem of knowing which item is selected based on the current location

And navigation could be made by using `useHistory` hook then history.push(path)

cypress/integration/app/sideBarNavigation.js
HellWolf93: This is not possible as cypress commands does not return Promises.
https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Commands-Are-Promises

packages/app/src/components/SideBarNavigation/index.js
HellWolf93: > also, I don't think we need to implement that logic our self here, take a look at https://reacttraining.com/react-router/web/api/Hooks/useroutematch ... `match` API will solve the problem of knowing which item is selected based on the current location

nice tip, thanks

packages/app/src/components/SideBarNavigation/index.js
```diff
@@ -0,0 +1,47 @@
...
+            name: element.props.name,
+            exact: element.props.exact,
+        };
+    });
+    const currentRoute = routes.find((route) => {
+        if (route.exact) {
+            return route.path === location.pathname;
+        }
+        return location.pathname.startsWith(route.path);
+    });
```
reiniergs, line 22: let's use useRouteMatch or match API here, there isn't a need to build our own logic for matching which is the selected item.

reiniergs: useRouteMatch('some/path') will return null it the current location does not match path provided.

Issue #38, opened 2020-05-25
feat: implement on RainbowFirebaseApp `initialize` prop


TahimiLeonBravo:
#### Requirements
En each project we make this
```function App() {
    const [isInitializing, setIsInitializing] = useState(true);
    useEffect(() => {
        (async () => {
            await initApp();
            setIsInitializing(false);
        })();
    }, []);
    return (
        <RainbowFirebaseApp app={firebase} theme={theme}>
            <RenderIf isTrue={!isInitializing}>
                <WhenAuthenticated path="/" redirect="/error">
                    <AuthRoutes />
                </WhenAuthenticated>
                <WhenNoAuthenticated path="/error">
                    <Error />
                </WhenNoAuthenticated>
            </RenderIf>
            <RenderIf isTrue={isInitializing}>
                <SpinnerContainer>
                    <Spinner />
                </SpinnerContainer>
            </RenderIf>
        </RainbowFirebaseApp>
    );
}
```

#### We want:

```const initialize = async (props) => {
   // props are the same passed into RainbowFirebaseApp    
}

<RainbowFirebaseApp initialize={initialize}>
   <WhenAuthenticated path="" />  
<RainbowFirebaseApp>
```

PR #44, opened 2020-05-27
feat: implement on RainbowFirebaseApp `initialize` prop

HellWolf93:
fix: #38


packages/app/src/components/App/index.js
```diff
@@ -57,6 +73,8 @@ RainbowFirebaseApp.propTypes = {
     translations: PropTypes.object,
     /** Object with your application Redux reducers. */
     reducers: PropTypes.object,
+    /** A funtion to initialize the app. AppSpiner will be visible while this function is running. */
```
maxxgreene, line 76: you have a typo `funtion` --> `function`. Also, we should specify it must be an async function

packages/app/src/components/App/index.js
```diff
@@ -66,6 +84,7 @@ RainbowFirebaseApp.defaultProps = {
     translations: {},
     children: null,
     reducers: {},
+    initialize: () => {},
```
maxxgreene, line 87: if there is a default value it means there a function that will execute by default, is that right?

packages/app/src/components/App/index.js
```diff
@@ -24,12 +26,26 @@ const RainbowFirebaseApp = (props) => {
...
+        async function initializeApp() {
+            if (initialize) {
+                setLoading(true);
+                await initialize();
+                setLoading(false);
+            }
+            setIsInitializing(false);
+        }
+        initializeApp();
+    }, [initialize]);
```
maxxgreene, line 39: it shouldn't depend on `initialize` prop, the `useEffect` callback function should be invoked once, the first time the component renders

TahimiLeonBravo:
@HellWolf93 please review the conflicts here

HellWolf93:
> @HellWolf93 please review the conflicts here

Fixed

PR #45, opened 2020-05-27
fix: `locale` prop on `RainbowFirebaseApp` component

blakmetall:
* method added inside libs/utils to retrieve browserLocale
* fallback to the browser locale when the locale prop is not passed


packages/app/src/libs/utils/getBrowserLocale.js
```diff
@@ -0,0 +1,8 @@
+export default function getBrowserLocale() {
+    return (
+        (navigator.languages && navigator.languages[0]) ||
+        navigator.language ||
+        navigator.userLanguage ||
+        'en-US'
+    );
+}
```
maxxgreene, line 8: please move this into `src/helpers` for now

PR #46, opened 2020-05-27
feat: add the rainbow logo to the icons module

blakmetall:
- icon moved (taken) from one location, then moved to icons folder
- paths updated
- before pull-request, conflicts needed to be solved, so a merge branch is there for that case


packages/app/docs/stories/sideBarNavigation.story.js
```diff
@@ -12,7 +12,7 @@ import {
 import RainbowFirebaseApp from '../../src/components/App';
 import SideBarNavigation from '../../src/components/SideBarNavigation';
 import SideBarOption from '../../src/components/SideBarOption';
-import RainbowLogo from './icons/rainbowLogo';
+import { RainbowLogo } from '../../../icons/src';
```
maxxgreene, line 15: it should be imported like
```js
import { rainbowIcon } from '@rainbow-modules/icons';
```

Issue #49, opened 2020-06-04
feat: allow custom spinner on RainbowFirebaseApp


TahimiLeonBravo:
### Stories design
- https://zpl.io/agYe9gD
- https://zpl.io/aR5roNN

reiniergs:
API Spec

### How to use custom spinner
```js
<RainbowFirebaseApp 
    spinner={<Spinner {...props} /> }> // ypu can pass you own spinner
</RainbowFirebaseApp>
```

### How to show app spinner
```js
import { showAppSpinner } from '@rainbow-modules/app';

showAppSpinner();
```

PR #51, opened 2020-06-11
feat: allow custom spinner on RainbowFirebaseApp

rgah2107:
fix: #49


packages/app/src/components/App/index.js
```diff
@@ -81,13 +98,15 @@ RainbowFirebaseApp.propTypes = {
...
     children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.object]),
+    spinnerChild: PropTypes.node,
 };
 
 RainbowFirebaseApp.defaultProps = {
     theme: undefined,
     locale: undefined,
     translations: {},
     children: null,
+    spinnerChild: null,
```
maxxgreene, line 109: spinner

packages/app/src/components/App/index.js
```diff
@@ -12,9 +12,19 @@ import { updateAppActions } from '../../actions';
...
+        locale,
+        translations,
+        children,
+        reducers,
+        initialize,
+        spinnerChild,
+    } = props;
     const firebaseContext = useMemo(() => ({ app }), [app]);
     const [isLoading, setLoading] = useState(false);
+    const [spinnerParams, setSpinnerParams] = useState({});
```
maxxgreene, line 27: why we we need a state for the spinner, by default the spinner has fix params values and if someone passes a custom spinner we should use that one
```
<RainbowFirebaseApp spinner={<MySpinner />}  >
...
```
```

Comment (rgah2107) [reply packages/app/src/components/App/index.js:27]: we cannot pass a custom spinner that way, because we would break the logic, we only have to pass the parameters of the spinner and its child in case it has any

Here is an example of a custom spinner: 
`export const appWithInitialize = () => {
    const initialize = async () => {
        showAppSpinner({
            size: 'x-small',
            variant: 'base',
        });
        await sleep(2000);
    };

    return (
        <RainbowFirebaseApp app={app} initialize={initialize}>
            <p>App is ready</p>
        </RainbowFirebaseApp>
    );
};`

or you can see the example of the spinner itself

Comment (maxxgreene) [reply packages/app/src/components/App/index.js:27]: `because it would break the logic` ?? can you elaborate more on this

Comment (maxxgreene) [reply packages/app/src/components/App/index.js:27]: This spinner renders at the app level and there is no use case for allowing spinner customization, only showAppSpinner() without params it's good for now.

Comment (rgah2107) [reply packages/app/src/components/App/index.js:27]: It seems that I misinterpret what is asked, let me check it before

File: packages/app/src/components/App/index.js [hunk 3]
```diff
@@ -51,7 +52,7 @@ const RainbowFirebaseApp = (props) => {
                         <RenderIf isTrue={!isInitializing}>
                             <BrowserRouter>{children}</BrowserRouter>
                         </RenderIf>
-                        <AppSpinner isLoading={isLoading} />
+                        <RenderIf isTrue={!!isLoading}>{currentSpinner}</RenderIf>
```
reiniergs, line 55: this value is internal and is always boolean, remove !!

packages/app/docs/stories/appSpinner.story.js
```diff
@@ -60,7 +76,34 @@ export const spinner = () => {
...
+    const customSpinner = (
+        <AppSpinner
+            type="arc"
+            size="x-large"
+            variant="brand"
+            label={<Label>loading react-rainbow</Label>}
+        >
+            <SpinnerLogo />
+        </AppSpinner>
+    );
```
reiniergs, line 89: move custom spinner out of the arcSpinnerWithChild function since does not depend on any internal scope

reiniergs:
LGTM, some minor changes

LeandroTorresSicilia:
there are issues with the spinner stories, when click the show spinner button in the second story then show the spinner for the first story

packages/app/docs/stories/appSpinner.story.js
```diff
@@ -4,6 +4,7 @@ import styled from 'styled-components';
 import { RainbowLogo } from '@rainbow-modules/icons';
 import app from '../../../../firebase';
 import RainbowFirebaseApp from '../../src/components/App';
+import AppSpinner from '../../src/components/AppSpinner';
 import { showAppSpinner, hideAppSpinner } from '../../src/actions';
```
TahimiLeonBravo, line 9: maintain this story like before and add a new story with the new feature

Issue #52, opened 2020-06-12
feat: add validation module


TahimiLeonBravo:
@rainbow-modules/validation

#### Design
https://zpl.io/bPGrjZp

#### Requirements
- [x] isEmail 
- [x] isDate
- [x] isDateInFuture
- [x] isDateInPast
- [x] isNumber
- [x] isPhoneNumber
- [x] isEmpty

NOTE: The validations should be similar to http://validatejs.org/

#### Usage
```import { isEmail } from '@rainbow-modules/validation';

isEmail('tahimi@nexxtway') // true```

PR #54, opened 2020-06-15
feat: add validation module

wildergd:
fix: #52 

### Changes
- add isNumber validator
- add isEmpty validator
- add isEmail validator
- add isDate validator 
- add isDateInPast validator
- add isDateInFuture validator 
- add isPhoneNumber validator 
- add isNull validator 
- add isUndefined validator 
- add isDefined validator


packages/validation/src/__test__/isDateInFuture.spec.js
```diff
@@ -0,0 +1,16 @@
+import isDateInFuture from '../isDateInFuture';
+
+describe('isDateInFuture', () => {
+    it('should return true', () => {
+        const values = [new Date(2021, 11, 23), '2020-12-19'];
```
reiniergs, line 5: these values are going to fail next year, make them computed based on today

packages/validation/src/__test__/isDate.spec.js
```diff
@@ -0,0 +1,16 @@
+import isDate from '../isDate';
+
+describe('isDate', () => {
+    it('should return true', () => {
+        const values = [new Date(), '2018-12-19', '2018/04/04', '2018/04/04 12:00:00 am'];
```
reiniergs, line 5: I don't know if we want to return true for string formatted like a valid date, isDate should return true only for Date instances

packages/validation/src/isDate.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default function isDate(value) {
+    if (isEmpty(value) || isNumber(value)) return false;
+
+    if (isNumber(Date.parse(value))) return true;
+
+    const date = new Date(value);
+    if (isNumber(date.getTime())) return true;
+
+    return false;
```
reiniergs, line 12:
```js
Object.prototype.toString.call(foo) === '[object Date]'
```

packages/validation/src/isEmail.js
```diff
@@ -0,0 +1,9 @@
+import isEmpty from './isEmpty';
+
+// eslint-disable-next-line no-useless-escape
+const emailRegEx = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+export default function isEmail(value) {
+    if (isEmpty(value)) return false;
```
reiniergs, line 7: why check that is empty if the regex will fail with empty values

packages/validation/src/isNullOrUndefined.js
```diff
@@ -0,0 +1,3 @@
+export default function isNullOrUndefined(value) {
+    return value === null || value === undefined;
```
reiniergs, line 2: we should separate these ones `isNull` & `isUndefined` also `isDefined`

packages/validation/src/isPhoneNumber.js
```diff
@@ -0,0 +1,107 @@
...
+    /^(\+?90|0)?5\d{9}$/,
+    /^(\+?38|8)?0\d{9}$/,
+    /^(\+?84|0)((3([2-9]))|(5([2689]))|(7([0|6-9]))|(8([1-6|89]))|(9([0-9])))([0-9]{7})$/,
+    /^((\+|00)86)?1([3568][0-9]|4[579]|6[67]|7[01235678]|9[189])[0-9]{8}$/,
+    /^(\+?886\-?|0)?9\d{8}$/,
+];
+
+export default function isPhoneNumber(value) {
+    if (isEmpty(value)) return false;
+    return matchers.reduce((hasMatched, matcher) => hasMatched || matcher.test(value), false);
```
reiniergs, line 106: use `.some` then you dont have to iterate the entire list 
```js
    return matchers.some(matcher => matcher.test(value));
```

reiniergs:
@wildergd @TahimiLeonBravo It looks good to me, minor changes ... but I don't see examples/docs :/

PR #57, opened 2020-06-17
feat: add more validations

wildergd:
fix: #57 and #58 

### Changes
- add isArray validator
- add isBoolean validator
- add isFunction validator 
- add isPromise validator
- add isInteger validator


packages/validation/src/isPromise.js
```diff
@@ -0,0 +1,5 @@
+import isFunction from './isFunction';
+
+export default function isPromise(value) {
+    return !!value && isFunction(value.then);
```
reiniergs, line 4:
```js
const fakePromise = { then: () => {}};
isPromise(fakePromise); // true 
```
Here the right solution
```js
const isPromise = (value) => {
    return Promise.resolve(value) === value
};
```

packages/validation/src/isAsync.js
```diff
@@ -0,0 +1,6 @@
+import isPromise from './isPromise';
+import isFunction from './isFunction';
+
+export default function isAsync(value) {
+    return isPromise(value) || (isFunction(value) && value() instanceof Promise);
+}
```
reiniergs, line 6: why execute the function to know if it's async, what happens if executing the function cause undesired results :/

we want to know if the function uses async/await
```js
return value.constructor.name === "AsyncFunction";
```

packages/validation/src/isArray.js
```diff
@@ -0,0 +1,3 @@
+export default function isArray(value) {
+    return Object.prototype.toString.call(value) === '[object Array]';
```
reiniergs, line 2: this is fine but we have Array.isArray the it does the same :)

wildergd: I made this validator for 100% compatibity with ie... as Array.isArray is only supported in ie 9+

wildergd: but it can be removed

packages/validation/src/isAsync.js
wildergd: this are the tests results with return value.constructor.name === "AsyncFunction"; method...
async function () {} -> false
async () => {} -> false,
new Promise(() => {})  -> false... // this can be fixed with isPromise check

packages/validation/docs/isPromise.story.mdx
```diff
@@ -0,0 +1,18 @@
...
+`isPromise` checks if the given value is a promise. This used the
+same semantics as the ECMAScript spec which means that any non empty
+object that has a .then function is a promise.
+
+```js
+isPromise(new Promise(function () {}));
+// returns true
+
+isPromise({ then: function () {} });
+// returns true
```
reiniergs, line 14: uumm, this not right :/

wildergd: already fixed

packages/validation/src/isAsync.js
maxxgreene: I understand but if I don't want to execute the function in any way ?? what if I'm checking if is async then execute another function

wildergd: checking if method is async is quite complicated... i found another solution that works without executing the function... i don't like it... it is based on string matching in the code generated by babel... code that  may change from one version to another... but it works

every function with `async` keyword is translated to the same code while normal functions  keep their original code
```js
console.log((async function () {}).toString().trim());
```
outputs:
```js
function () {
    var self = this,
    args = arguments;
    return new Promise(function (resolve, reject) 
        var gen = fn.apply(self, args);
        function _next(value) {
              asyncGeneratorStep(gen, resolve, reject, _next, _throw, "next", value);
        }
        function _throw(err) {
            asyncGeneratorStep(gen, resolve, reject, _next, _throw, "throw", err);
        }
        _next(undefined);
    });
}
```

packages/validation/src/isAsync.js
```diff
@@ -0,0 +1,9 @@
+import isPromise from './isPromise';
+import isFunction from './isFunction';
+
+export default function isAsync(value) {
+    return (
+        isPromise(value) ||
+        (isFunction(value) && /return new Promise|asyncGeneratorStep/.test(value.toString().trim()))
```
maxxgreene, line 7:
```js
const AsyncFunction = (async () => {}).constructor;
const GeneratorFunction = (function* () => {}).constructor;

const isAsync = (value) => {
return isPromise(value) || 
    (isFunction(value) && value.constructor.name === 'AsyncFunction')  ||  // native 
    (value instanceof AsyncFunction && AsyncFunction !== Function && AsyncFunction !== GeneratorFunction);
};
```

wildergd: false for every value....

maxxgreene: you have to include `isFunction(value) && value.constructor.name === 'AsyncFunction'` because this function wont run always as a compiled version by babel, in fact in nodejs will return false if you don't include it

Issue #82, opened 2020-08-13
fix: improve icons module


TahimiLeonBravo:
### Requirements
- [ ] make the dropdown icon info only show when the icon is selected
- [ ] render the icons module on the "MODULES" section
<img width="1439" alt="Screen Shot 2020-08-13 at 5 51 01 PM" src="https://user-images.githubusercontent.com/32083556/90194876-d44a7c80-dd8d-11ea-8596-f2b7dc8caea8.png">

- [ ] change the "error" and "done" icons

PR #83, opened 2020-08-17
fix: improve icons module

rgah2107:
fix: #82

### Changes
- make the dropdown icon info only show when the icon is selected
- render the icons module on the "MODULES" section
- change the "error" and "done" icons


packages/icons/docs/stories/rainbowIconItem.js
```diff
@@ -80,7 +80,6 @@ const RainbowIconItem = (props) => {
             <StyledButton
                 id={iconBoxId}
                 ref={triggerRef}
-                onMouseEnter={openOverlay}
```
yvmunayev, line 83: You have to check this component well and eliminate all the necessary logic, you only have to leave what is related to the isClickTooltip

yvmunayev: @LeandroTorresSicilia this component is based on the HelpText, if we change the behavior, everything related to the hover and the setTimeout is not necessary.

Issue #86, opened 2020-08-25
fix: add `rainbow-module` logo


TahimiLeonBravo:
### Requirements
- [ ] Add the logo and name to the sidebar header
https://zpl.io/amkrJ19
- [ ] add the logo on the favicon

PR #87, opened 2020-08-27
fix: add rainbow-module logo

rgah2107:
fix: #86

### Changes
- add the logo and name to the sidebar header
- add the logo on the favicon


rgah2107:
the image in the header of the sidebar may change...

PR #89, opened 2020-08-28
feat: implement "showNotification" component

HellWolf93:


packages/app/src/components/AppNotificationManager/index.js
```diff
@@ -0,0 +1,55 @@
...
+            }, timeout + 600);
+        }
+    };
+
+    useEffect(() => {
+        updateAppActions({
+            addNotification,
+            removeNotification,
+        });
+    });
```
yvmunayev, line 41: if you convert addNotification and removeNotification to useCallback this effect can only be done once.

packages/app/src/components/AppNotificationManager/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+    const removeNotification = (notification) => {
+        setNotifications((notifications) => {
+            return notifications.filter((item) => {
+                return item.notification !== notification;
+            });
+        });
+    };
+
+    const addNotification = (notification, timeout) => {
```
yvmunayev, line 19: useCallback

packages/app/src/components/AppNotificationManager/index.js
```diff
@@ -0,0 +1,55 @@
...
+import { createPortal } from 'react-dom';
+import { updateAppActions } from '../../actions';
+import { StyledNotificationContainer } from './styled';
+import Notification from './notification';
+import uniqueId from './utils/uniqueId';
+
+const AppNotificationManager = () => {
+    const [notifications, setNotifications] = useState([]);
+
+    const removeNotification = (notification) => {
```
yvmunayev, line 11: useCallback

packages/app/src/components/AppNotificationManager/notification.js
```diff
@@ -0,0 +1,43 @@
...
+        setTimeout(() => {
+            setShouldExit(true);
+        }, timeout);
+    }
+
+    if (contentMeta) {
+        const { height } = contentMeta;
+        setTimeout(() => {
+            setShown(true);
+        }, 0);
```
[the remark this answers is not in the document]
HellWolf93: The idea behind the show status is to update the component after it has rendered the first time, hence the timeout. This is necessary because we need to render the notification outside of the view, and then apply change the shown state, so the StyledComponent updates the styles and runs the transitions. That is why after the content meta is resolved, the notification is rendered with shown=false, and then updates to shown=true

packages/app/src/components/AppNotificationManager/index.js
HellWolf93: Adding `useCallback` here does not have any benefits, I ran the profiler on the component and rendering times are bigger when using `useCallback`

packages/app/src/components/AppNotificationManager/index.js
HellWolf93: same ^^

packages/app/src/components/AppNotificationManager/index.js
HellWolf93: I ran performance tests on the component with and without `useCallback` and it turned out that is faster without it, so I think that even if we update the app actions object on each render still is more efficient that calling `useCallback`

packages/app/src/components/AppNotificationManager/index.js
HellWolf93: same ^^

packages/app/docs/stories/appNotification.story.js
```diff
@@ -0,0 +1,121 @@
...
+export const showAppLevelNotification = () => {
+    const showNotification = () => {
+        showAppNotification(
+            {
+                title: 'Notification',
+                description: 'This is the notification description.',
+                icon: 'success',
+            },
+            null,
+        );
```
LeandroTorresSicilia, line 61: here we need to use it like:
```
showAppNotification({
                title: 'Notification',
                description: 'This is the notification description.',
                icon: 'success',
                timeout: 2000,
})
```
in order to be more explicit with the timeout

packages/app/src/components/AppNotificationManager/index.js
```diff
@@ -0,0 +1,52 @@
...
+        setNotifications(notifications.filter((item) => item.key !== key));
+    };
+
+    const addNotification = (notification, timeout = 2000) => {
+        const key = uniqueId('notification');
+        setNotifications([{ key, timeout, ...notification }, ...notifications]);
+        if (timeout && timeout > 0) {
+            setTimeout(() => {
+                removeNotification(key);
+            }, timeout + 600);
```
LeandroTorresSicilia, line 21: why we need use `timeout + 600`?

packages/app/src/components/AppNotificationManager/index.js
```diff
@@ -0,0 +1,52 @@
...
+        }
+    };
+
+    useEffect(() => {
+        updateAppActions({
+            addNotification,
+        });
+    });
```
LeandroTorresSicilia, line 29: this is meant to run on each rerender or only the first time?

packages/app/src/components/AppNotificationManager/notification.js
```diff
@@ -0,0 +1,85 @@
...
+    onRequestClose: PropTypes.func,
+    timeout: PropTypes.number,
+};
+
+AppNotification.defaultProps = {
+    icon: null,
+    title: null,
+    description: null,
+    onRequestClose: () => {},
+    timeout: undefined,
```
LeandroTorresSicilia, line 82: the notification should have a default value of 2000ms

packages/app/src/components/AppNotificationManager/styled.js
```diff
@@ -0,0 +1,142 @@
...
+        transition-duration: 0.2s, 0.4s, 0.4s, 0s;
+        transition-delay: 0.4s, 0s, 0s, 0.6s;
+        transform: translateX(100%);
+        height: 0;
+        `}
+`;
+
+/*
+export const StyledNotification = styled.div`
+    @keyframes enter {
```
LeandroTorresSicilia, line 78: remove commented code if not necessary

TahimiLeonBravo (approved):

packages/app/docs/stories/appNotification.story.js
HellWolf93: Done

packages/app/src/components/AppNotificationManager/index.js
HellWolf93: It is necessary in order to give the transition enough time to finish, this could be done in 2 ways, starting the transition on timeout and wait 600ms (duration of the transition) to remove the component from the tree, or start the transition 600ms before the timeout and remove the component from the tree on timeout, I went with the first option, but this can be changed easily

packages/app/src/components/AppNotificationManager/notification.js
HellWolf93: Done

packages/app/src/components/AppNotificationManager/styled.js
HellWolf93: Done

LeandroTorresSicilia (approved):

PR #107, opened 2020-09-16
feat: implement `DatePickerCarousel` component

### LINKED ISSUE #102: feat: implement `<DatePickerCarousel />` component on `@rainbow-modules/datetime` module
### Specs
- [ ] Implement the component functionality
- [ ] implement keyboard navigation
- [ ] add tests
- [ ] add documentation
- [ ] add real use-case example

### Design
- https://zpl.io/a71GkGW
- https://zpl.io/2yr1xLo (CalendarModal)

### Requirements
- [ ] It should have the same props as `DatePicker` component
- [ ] It should have a width of 100% on the container
- [ ] the day cards have a fixed width
- [ ] the day cards amount are calculated based on the width of the container on initial render
- [ ] moving cards when clicking on the arrows left/right happen 1 card at the time
- [ ] adding a new card to the carousel imply removing another card (e.g. if the user want to move to the right then one card is added into the right and one card is removed in the left side)
- [ ] resize of the container won't take effect on the component since only visible card in the container are shown with proper CSS
- [ ] the selected card should move when the customer uses the keyboard (we need to implement the keyboard navigation)
- [ ] it should open a calendarModal when the customer click on the header content
- [ ] It should maintain real-time sync the date selected on the carousel with the date shown on the header

wildergd:
Fix #102 

- implement DatePickerCarousel component


packages/datetime/src/components/DatePickerCarousel/carouselCalendar/hooks/useWindowResize.js
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useCallback } from 'react';
+
+export default function useWindowResize(handler, isListening = true) {
```
reiniergs, line 3: this a hook we will use in other components, I think we should exposed it if we are not :)

reiniergs: also, what we are looking for is resize of the container where the component was rendered, that could happen without the window being resized. I know the Table component implements something to detect changes of size in a container.

packages/datetime/src/components/DatePickerCarousel/carouselCalendar/icons/rightArrow.js
```diff
@@ -0,0 +1,17 @@
...
+                    <g transform="translate(368.000000, 3.000000)">
+                        <g transform="translate(55.000000, 14.000000)">
+                            <path d="M0.876459893,13.9508128 C0.671251337,13.9508128 0.465219251,13.8726524 0.308,13.7150588 C-0.00598930481,13.4010695 -0.00598930481,12.8929519 0.308,12.5789626 L5.87445989,7.01287701 L0.308,1.44641711 C-0.00598930481,1.13242781 -0.00598930481,0.623860963 0.308,0.310320856 C0.621989305,-0.0036684492 1.13055615,-0.0036684492 1.44447059,0.310320856 L7.57856684,6.44441711 C7.89255615,6.75840642 7.89255615,7.26697326 7.57856684,7.58051337 L1.44447059,13.714984 C1.2877754,13.8722032 1.08211765,13.9508128 0.876459893,13.9508128 Z" />
+                        </g>
+                    </g>
+                </g>
+            </g>
+        </svg>
+    );
+}
```
reiniergs, line 17: it's fine is the component implement its own icons, but we should consider using the ones on `icons` module

packages/datetime/src/components/DatePickerCarousel/carouselCalendar/hooks/useWindowResize.js
wildergd: ok i will check... i think i can make a hook, similar to uswWindowResize, that can be useContainerResize... that monitors a Ref and checks if ref.current.clientWidth or ref.current.clientHeight changed

wildergd: also.. we have to check hooks in react-rainbow components that should be exposed as well in order to build components outside the library... like useLocale, useReduxForm.... etc

packages/datetime/src/components/DatePickerCarousel/carouselCalendar/icons/rightArrow.js
wildergd: ok.. i will check if they are already in `icons` module and move them if they are not

TahimiLeonBravo (approved):

packages/datetime/src/components/DatePickerCarousel/index.js
```diff
@@ -0,0 +1,200 @@
...
+
+const DatePickerCarousel = React.forwardRef((props, ref) => {
+    const {
+        value,
+        minDate,
+        maxDate,
+        placeholder,
+        onClick,
+        onChange,
+        // onFocus,
```
TahimiLeonBravo, line 18: we should remove the not used props

TahimiLeonBravo (approved):

Issue #103, opened 2020-09-14
feat: implement `<ColoredStatusColumn />` component on `@rainbow-modules/listView` module


TahimiLeonBravo:
#### Specs
- [ ] Implement component
- [ ] Add documentation
- [ ] Add test
- [ ] add real use cases examples

#### Design
https://zpl.io/Vx7ny83

#### Requirements
- [ ] the text color must be calculated according to the `background-color` of the cell
- [ ] the default colors should be the theme colors

#### Props
`colors`: Map by default contains Rainbow Status colors
`textTransform`: ‘capitalize | uppercase | lowercase’
`value` // status comes in the value
`row` // unusable 

// by default
const colors = {
   'success': '',
   'error': '',
   'warning': '', 

#### example 1
```
const data = [
   { status: 'completed', ... },
   { status: 'in progress', ... },
   { status: 'cancelled', ... },
];

const colors = {
    completed: { backgroundColor: '#f00', color: 'rgba(255, 0, 0, 0.8)' },
    cancelled: '#0f0',
    'in progress': '#00f',
};

const CustomColoredStatusColumn = (...props) => <ColoredStatusColumn ...props colors={colors} />;

<Table data={data}>
   <Column filed="status" component={CustomColoredStatusColumn} />
</Table>
```

#### example 2
```
const data = [
   { status: 'completed', ... },
   { status: 'in progress', ... },
   { status: 'cancelled', ... },
];

<Table data={data}>
   <Column field="status" component={ColoredStatusColumn}
</Table>

// by default
const colors = {
   'success': '',
   'error': '',
   'warning': '', 
}
```

PR #109, opened 2020-09-17
feat: implement coloredstatuscolumn component on rainbow-modules

amontalvof:
fix: #103


packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -0,0 +1,59 @@
...
+import PropTypes from 'prop-types';
+import useTheme from 'react-rainbow-components/libs/hooks/useTheme.js';
+import getContrastText from 'react-rainbow-components/styles/helpers/color/getContrastText.js';
+import { StyledContainer } from './styled';
+
+const ColoredStatusColumn = ({ value, colors, textTransform }) => {
+    const theme = useTheme();
+
+    const { success, warning, error } = theme.rainbow.palette;
+    const defaultColors = {
```
TahimiLeonBravo, line 11: @reiniergs I think we should review the default map

TahimiLeonBravo (changes requested):

packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -0,0 +1,59 @@
...
+    /** An object with the mapping that indicates the backgroundColor and color of the component */
+    colors: PropTypes.object,
+    /** A string that indicates the format of the font capitalize | uppercase | lowercase */
+    textTransform: PropTypes.string,
+};
+
+ColoredStatusColumn.defaultProps = {
+    value: undefined,
+    colors: undefined,
+    textTransform: undefined,
```
TahimiLeonBravo, line 58: the default value on `textTransform` should be `capitalize`

TahimiLeonBravo (changes requested):

packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+export default ColoredStatusColumn;
+
+ColoredStatusColumn.propTypes = {
+    /**  */
+    value: PropTypes.node,
+    /** An object with the mapping that indicates the backgroundColor and color of the component */
+    colors: PropTypes.object,
+    /** A string that indicates the format of the font capitalize | uppercase | lowercase */
+    textTransform: PropTypes.string,
```
TahimiLeonBravo, line 52: I don't see the `row` prop

TahimiLeonBravo (changes requested):

packages/listview/src/components/ColoredStatusColumn/index.js
TahimiLeonBravo: we are not going to use the row prop

packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -0,0 +1,59 @@
...
+            background = map[value].backgroundColor;
+        } else {
+            background = map[value];
+        }
+        if (map[value].color) {
+            color = map[value].color;
+        } else {
+            color = getContrastText(background);
+        }
+    }
```
reiniergs, line 35: move this logic into functions that you can test and it will simplify the complexity e.g.

```js
const backgroundColor = getBackgrondColor(value);
const color = getColor(value);

const getBackgroundColor = (value) => {
    if (typeof map[value] === 'string' && isValidColor(map[value])) {
        return map[value];
    } 
    If (map[value] && typeof map[value].backgroundColor === 'string' && isValidColor(map[value].background)) {
        return map[value].backgroundColor;
    }
    return defaultBackgroundColor; // white
}

const getColor(value) => {
    if (map[value] && typeof map[value].color === 'string' && isValidColor(map[value].color)) {
       return map[value].color;
    }
    const backgroundColor = getBackgroundColor(value);
    return getContrastText(backgroundColor);
}
```

packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -0,0 +1,59 @@
...
+    const { success, warning, error } = theme.rainbow.palette;
+    const defaultColors = {
+        success: success.main,
+        warning: warning.main,
+        error: error.main,
+    };
+
+    const map = {
+        ...defaultColors,
+        ...colors,
```
reiniergs, line 19: what happens is the colors is undefined and you try to use spread
```
Uncaught ReferenceError: colors is not defined
```

reiniergs: also, you are teaching new colors, no completely overriding the colors

reiniergs: ... and the components shouldn't be case sensitive, `Success` should be the same as `success`

packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -0,0 +1,49 @@
...
+        >
+            {value}
+        </StyledContainer>
+    );
+};
+
+export default ColoredStatusColumn;
+
+ColoredStatusColumn.propTypes = {
+    /**  */
```
TahimiLeonBravo, line 29: add the description for the value prop

TahimiLeonBravo (changes requested):

maxxgreene (approved):
+1

Issue #116, opened 2020-09-21
feat: improve DatePickerCarousel component


TahimiLeonBravo:
### Specs
- [x] Add unit tests
- [x] Add integration tests
- [x] Add documentation
- [ ] Add a meaningful example for real use case

### Requirements
-  [ ] add specs proposal
- [x] It should have a "Components" section under the "Datetime" module
- [x] It should have the "DatePickerCarousel" component doc on the "Components" section

PR #117, opened 2020-09-21
test: add tests for `DatePickerCarousel` component

wildergd:
fix: #116 

## Changes proposed in this PR:
- test: add unit test DatePickerCarousel component
- fix: minor improvements


packages/datetime/docs/stories/DatePickerCarousel.story.js
```diff
@@ -11,12 +11,12 @@ const Container = styled.div`
 `;
 
 // eslint-disable-next-line react/prop-types
-const PickDate = ({ locale }) => {
-    const [date, setDate] = useState(new Date());
-    return <DatePickerCarousel locale={locale} value={date} onChange={(date) => setDate(date)} />;
+const PickDate = (props) => {
+    const [date, setDate] = useState(new Date('2020/09/15'));
+    return <DatePickerCarousel {...props} value={date} onChange={(date) => setDate(date)} />;
```
LeandroTorresSicilia, line 16: here you can
```onChange={setDate}```

File: packages/datetime/src/components/DatePickerCarousel/carouselCalendar/__test__/carouselCalendar.spec.js [hunk 2]
```diff
@@ -0,0 +1,52 @@
...
+    });
+    it('should set to disable the scroll right button when maxDate is equal or below than the last visible date', () => {
+        const date = new Date('2020/10/21');
+        const maxDate = new Date('2020/10/22');
+        const component = mountWithApplication(<CarouselCalendar value={date} maxDate={maxDate} />);
+        const scrollLeft = component.find('ButtonIcon').at(1);
+        expect(scrollLeft.prop('disabled')).toBe(true);
+    });
+
+    it.skip('should render the right date when value is updated', () => {
```
LeandroTorresSicilia, line 43: let's try no skip tests

packages/datetime/src/components/DatePickerCarousel/carouselCalendar/__test__/dayCard.spec.js
```diff
@@ -0,0 +1,25 @@
...
+        const component = mountWithApplication(<DayCard date={date} />);
+        expect(component.find('span').at(0).text()).toBe('Wed');
+        expect(component.find('span').at(1).text()).toBe('21');
+    });
+    it('should call onChange with the right date when click the button', () => {
+        const date = new Date('04/24/2019');
+        const onChangeMockFn = jest.fn();
+        const component = mountWithApplication(<DayCard date={date} onChange={onChangeMockFn} />);
+        component.find('button').simulate('click');
+        expect(onChangeMockFn).toHaveBeenCalledWith(new Date('04/24/2019'));
```
LeandroTorresSicilia, line 18: this test is repeated in the above three test files, I think we should keep only the first one in the main component since it is the final result we want

packages/datetime/src/helpers/mountWithApplication.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import { Application } from 'react-rainbow-components';
+
+export default function mountWithApplication(el, options) {
+    return mount(<Application>{el}</Application>, options).find(el.type);
+}
```
LeandroTorresSicilia, line 7: it more readable when use `element` instead of el, in react-rainbow we are updating and adding eslint rules and now it will not let you add variable names with less than 3 characters

packages/datetime/src/components/DatePickerCarousel/carouselCalendar/__test__/carouselCalendar.spec.js
wildergd: i set this as .skip because it is a test i want to make but `setProps` method  fails with forwardRef components.. 
it fires this exception: 
  Warning: Failed prop type: Invalid prop `value` supplied to `ForwardRef`. in ForwardRef

packages/datetime/src/components/DatePickerCarousel/carouselInput/__test__/carouselInput.spec.js
```diff
@@ -0,0 +1,24 @@
...
+import CarouselCalendar from '../../carouselCalendar';
+
+jest.mock('../../carouselCalendar/helpers/getVisibleCards', () => jest.fn(() => 7));
+
+describe('<CarouselInput />', () => {
+    it('should pass right props to child components', () => {
+        const date = new Date('2020/10/21');
+        const component = mountWithApplication(<CarouselInput value={date} />);
+        expect(component.find(CarouselCalendar).prop('value')).toEqual(date);
+        expect(component.find(CarouselCalendar).prop('value')).toEqual(date);
```
LeandroTorresSicilia, line 13: this expect is twice

PR #145, opened 2020-10-04
feat: added delete on imagegallery component

amontalvof:
#140


packages/storage/src/components/ImageGallery/image.js
```diff
@@ -1,19 +1,25 @@
...
     StyledContainerSpinner,
     StyledImage,
+    StyledDeleteButton,
+    StyledCheckContainer,
+    TrashIcon,
 } from './styled';
 
 export default function Image(props) {
-    const { imageRef, onSelect, onError } = props;
+    const { imageRef, onSelect, onError, allowDelete, handleDeleteClick } = props;
```
LeandroTorresSicilia, line 19: when passing a callback function as a prop we use to start its name with on... like onDelete, the function that handles the callback is the one that is called like this with handle...

packages/storage/src/components/ImageGallery/image.js
```diff
@@ -28,16 +34,52 @@ export default function Image(props) {
...
+            <StyledContainerImage type="button" onClick={handleClick}>
                 <RenderIf isTrue={!!src}>
                     <StyledImage
                         src={src}
                         alt={imageRef.name}
                         onLoad={() => setLoading(false)}
                         $loading={loading}
                     />
+                    <RenderIf isTrue={allowDelete}>
+                        <StyledDeleteButton>
```
LeandroTorresSicilia, line 70: why use an extra wrapper here?

packages/storage/src/components/ImageGallery/image.js
```diff
@@ -28,16 +34,52 @@ export default function Image(props) {
         })();
     }, [imageRef, onError]);
 
+    const handleClick = (e) => {
+        const isClickDeleteBtn = deleteBtn.current.htmlElementRef.current.contains(e.target);
```
LeandroTorresSicilia, line 38: isDeleteButtonClicked

packages/storage/src/components/ImageGallery/image.js
```diff
@@ -28,16 +34,52 @@ export default function Image(props) {
...
+                            <ButtonIcon
+                                variant="inverse"
+                                size="medium"
+                                icon={<Trash />}
+                                onClick={openModalDelete}
+                                ref={deleteBtn}
+                            />
+                        </StyledDeleteButton>
+                    </RenderIf>
+                    <StyledCheckContainer>
```
LeandroTorresSicilia, line 80: same ^^ when possible we need to avoid extra wrappers, we can pass the icon through a styled and apply necessary styles

LeandroTorresSicilia (approved):

TahimiLeonBravo (approved):

PR #148, opened 2020-10-06
feat: add `middlewares` prop to App and ReduxContainer components in `@rainbow-modules/app`

wildergd:
## Changes proposed in this PR:
- add `middlewares` prop to App and ReduxContainer components in `@rainbow-modules/app`


packages/app/docs/stories/appWithReduxMiddleware.story.js
```diff
@@ -0,0 +1,169 @@
...
+            store.dispatch({ type: 'CONNECTED' });
+        };
+
+        const onSocketClose = (store) => () => {
+            store.dispatch({ type: 'DISCONNECTED' });
+        };
+
+        const onError = (store) => () => {
+            store.dispatch({ type: 'DISCONNECTED' });
+        };
```
LeandroTorresSicilia, line 101: while staying connected if I switch off the internet of my PC the UI remains as connected, I think it should react since the socket connection should know there is not internet

LeandroTorresSicilia (approved):

Issue #146, opened 2020-10-04
feat: implement UniversalPicker


TahimiLeonBravo:
<UniversalPicker />

#### Usage:
```
<UniversalPicker {...samePropsVisualPicker} >
   <UniversalItem component={Item} {...samePropsVisualPickerItem}>
</UniversalPicker>

const Item = (props) => {
    const { state, disabled } = props;
    if (state.isSelected) {
        return </>
    }  
};
```

PR #156, opened 2020-10-11
feat: implement UniversalPicker

yvmunayev:
fix: #146


packages/forms/src/components/UniversalPicker/index.js
```diff
@@ -0,0 +1,119 @@
...
+            </StyledOptionsContainer>
+            <RenderIf isTrue={!!error}>
+                <StyledError id={errorMessageId}>{error}</StyledError>
+            </RenderIf>
+        </StyledContainer>
+    );
+}
+
+UniversalPicker.propTypes = {
+    /** The name of VisualPicker. */
```
LeandroTorresSicilia, line 76: The name of the UniversalPicker

packages/forms/src/components/UniversalPicker/index.js
```diff
@@ -0,0 +1,119 @@
...
+    name: PropTypes.string,
+    /** The value of the element. */
+    value: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
+    /** If true then a multiple selection is allowed */
+    multiple: PropTypes.bool,
+    /** The action triggered when a value attribute changes. */
+    onChange: PropTypes.func,
+    /** If is set to true the VisualPicker is required. This value defaults to false. */
```
LeandroTorresSicilia, line 84: same ^^ VisualPicker => UniversalPicker

packages/forms/src/components/UniversalPicker/index.js
```diff
@@ -0,0 +1,119 @@
...
+    required: PropTypes.bool,
+    /** The title at the top of the VisualPicker component. */
```
LeandroTorresSicilia, line 86: same ^^

packages/forms/src/components/UniversalPicker/index.js
```diff
@@ -0,0 +1,119 @@
...
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Specifies that an VisualPicker must be filled out before submitting the form. */
```
LeandroTorresSicilia, line 88: same ^^

packages/forms/src/components/UniversalPicker/index.js
```diff
@@ -0,0 +1,119 @@
...
+    error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** The size of the VisualPicker. Valid values are small, medium, and large.
```
LeandroTorresSicilia, line 90: same ^^

packages/forms/docs/stories/UniversalPicker.story.js
```diff
@@ -0,0 +1,414 @@
...
+        </StyledRadioItem>
+    );
+};
+
+const StyledRadioPicker = styled(UniversalPicker)`
+    width: 500px;
+
+    > div {
+        flex-direction: column;
+    }
```
LeandroTorresSicilia, line 321: we need to find another way to do this, maybe another prop variant, I will discuss it with @maxxgreene

packages/forms/src/components/UniversalPickerOption/__test__/universalPickerOption.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        const onChangeMock = jest.fn();
+        const component = mount(
+            <UniversalPickerOption name="option1" value={value} onChange={onChangeMock} />,
+        );
+        component
+            .find('input')
+            .first()
+            .simulate('change', { target: { checked: true } });
+
+        expect(onChangeMock).toBeCalledWith('option1', true);
```
LeandroTorresSicilia, line 45: I think we should remove this test since it is already testes in the parent component

packages/forms/src/components/UniversalPickerOption/index.js
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * A UniversalPickerItem.
+ */
+export default function UniversalPickerOption(props) {
+    return (
+        <UniversalPickerConsumer>
+            {(context) => <PickerOption {...props} {...context} />}
+        </UniversalPickerConsumer>
+    );
+}
```
LeandroTorresSicilia, line 67: since we are using functional components with hooks let's use the useContext hook instead of this wrapper

packages/forms/src/components/UniversalPickerOption/index.js
```diff
@@ -0,0 +1,99 @@
...
+                onFocus={() => setFocused(true)}
+                onBlur={() => setFocused(false)}
+            />
+
+            <StyledLabel
+                htmlFor={inputId}
+                onMouseEnter={() => setHover(true)}
+                onMouseLeave={() => setHover(false)}
+            >
+                <Item {...rest} state={state} />
```
LeandroTorresSicilia, line 52: let's send the primitive state's values directly instead of the state object like:
```
<Item {...rest} isSelected={isSelected} isFocused={isFocused} isHover={isHover} />
```

packages/forms/docs/stories/UniversalPicker.story.js
LeandroTorresSicilia: We discussed it already, we will add a `direction` prop that will manage this:
direction='horizontal' | 'vertical'
default value: 'horizontal'

TahimiLeonBravo:
@yvmunayev let's use the same example in the visual picker that we have in the library https://react-rainbow.io/#!/VisualPicker/1

LeandroTorresSicilia (approved):

Issue #167, opened 2020-10-20
test: `Mapbox` component


TahimiLeonBravo:
#### requirements
 - [ ] add unit test
 - [ ] add integration tests is needed

###### map

- [ ] should render two p with right text when accessToken is undefined
- [ ] should initially render the Spinner when accessToken is passed
- [ ] should init map with the center passed
- [ ] should call navigator.geolocation.getCurrentPosition when center is not passed
- [ ] should init map with resolved position
- [ ] should init map with the default center passed
- [ ] should render the children passed

PR #197, opened 2020-11-06
fix: test Mapbox component first commit

rgah2107:
fix: #167

### Changes
- test Mapbox component


setupTests.js
```diff
@@ -20,3 +20,15 @@ jest.mock('react-intl', () => {
...
+    GeolocateControl: jest.fn(),
+    Map: jest.fn(() => ({
+        addControl: jest.fn(),
+        on: jest.fn(),
+        remove: jest.fn(),
+        once: jest.fn(),
+    })),
+    NavigationControl: jest.fn(),
+}));
+jest.mock('mapbox-gl/dist/mapbox-gl.css', () => ({}));
```
LeandroTorresSicilia, line 34: I think this can go in the map test file since this will not be used across all components and modules like `react-intl`, thus you can take more control on what you want to mock

packages/mapbox/src/components/Map/__test__/map.spec.js
```diff
@@ -0,0 +1,44 @@
...
+jest.mock('mapbox-gl/dist/mapbox-gl', () => ({
+    GeolocateControl: jest.fn(),
+    Map: jest.fn(() => ({
+        addControl: jest.fn(),
+        on: jest.fn(),
+        remove: jest.fn(),
+        once: jest.fn(),
+    })),
+    NavigationControl: jest.fn(),
+}));
```
LeandroTorresSicilia, line 15: we should add more test cases:
- should init map with the center passed (when pass a valid center new mapboxgl.Map should be called with the right one)
- should call navigator.geolocation.getCurrentPosition when center is not passed (here if we can mock the success and error callbacks we should test both cases, when new mapboxgl.Map is called with the resolved position and when called with the defaultCenter passed)

LeandroTorresSicilia (approved):
Good job!!!

PR #202, opened 2020-11-17
Floating search button icon

amontalvof:


packages/forms/src/components/FloatingSearchButtonIcon/index.js
```diff
@@ -0,0 +1,143 @@
...
+        className,
+        disabled,
+        icon,
+        id,
+        shaded,
+        size,
+        style,
+        tabIndex,
+        title,
+        type,
```
LeandroTorresSicilia, line 22: type does not make sense here since it will be type="button" always, there are no use cases for submit and reset

packages/forms/src/components/FloatingSearchButtonIcon/index.js
```diff
@@ -0,0 +1,143 @@
...
+        variant,
+        placeholder,
+        onChange,
+        value,
+    } = props;
+
+    const triggerRef = useRef(null);
```
LeandroTorresSicilia, line 29: no null

LeandroTorresSicilia:
packages/forms/src/components/FloatingSearchButtonIcon/style.js is empty

LeandroTorresSicilia (approved):

Issue #212, opened 2020-11-19
feat: implement `<RecordHeader />` component on `@rainbow-modules/record`


TahimiLeonBravo:
#### Design
https://zpl.io/aX004ex

#### Usage
```
<RecordHeader />

usage:

<RecordHeader 
   label=""
   description=""
   icon={<IconFoo />}>
   actions={
      <ButtonGroup>
        <Button label="foo" onClick={} />
        <ButtonIcon icon={} onClick={} />
      </ButtonGroup>
   }
   tags={
      <>
       <Badge label="" />
       <Badge label="" />
      </> 
   }
</RecordHeader>
```

PR #214, opened 2020-11-19
feat: implemented recordheader module

amontalvof:
fix: #212


LeandroTorresSicilia:
@amontalvof there are conflicts here

packages/record/src/index.js
```diff
@@ -1,2 +1,3 @@
 // eslint-disable-next-line import/prefer-default-export
```
LeandroTorresSicilia, line 1: here we can remove this eslint rule exception since it is only needed when there is only one named export

LeandroTorresSicilia (approved):

Issue #209, opened 2020-11-19
feat: implement `RecordDetails` component on `@rainbow-modules/record`


TahimiLeonBravo:
#### Usage
```
<RecordDetails>
   <RecordField label="Name" value={user.name} type="text" />
</RecordDetails>
```

Issue #208, opened 2020-11-19
feat: implement `RecordPrimaryDetails` component on `@rainbow-modules/record`


TahimiLeonBravo:
#### Usage
```
<RecordPrimaryDetails>
   <RecordField
        label="Name"
        value={user.name}
        type="text" />

   <RecordField
        label="Bank Account"
        value={user.account}
        type="currency" />

</RecordPrimaryDetails>
```

#### Design
https://zpl.io/adZZko7

PR #215, opened 2020-11-19
feat: implemented recordprimarydetails module

amontalvof:
fix: #208 
fix: #209


LeandroTorresSicilia:
@amontalvof conflicts

packages/record/src/components/RecordDetails/index.js
```diff
@@ -0,0 +1,34 @@
...
+import PropTypes from 'prop-types';
+import { Provider } from '../../context';
+import { StyledContainer } from './styled';
+
+export default function RecordDetails(props) {
+    const { style, id, children, className } = props;
+
+    const context = {
+        privateVariant: 'horizontal',
+    };
```
LeandroTorresSicilia, line 11: this can be outside the component

packages/record/src/context.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+
+export const RecordFieldContext = React.createContext();
```
LeandroTorresSicilia, line 3: let's change this name since it coupling it to the RecordField, we can call it simply Context

packages/record/src/components/RecordPrimaryDetails/index.js
```diff
@@ -0,0 +1,34 @@
...
+import PropTypes from 'prop-types';
+import { Provider } from '../../context';
+import { StyledContainer } from './styled';
+
+export default function RecordPrimaryDetails(props) {
+    const { style, id, children, className } = props;
+
+    const context = {
+        privateVariant: 'border-vertical',
+    };
```
LeandroTorresSicilia, line 11: same ^^ use outside

LeandroTorresSicilia (approved):

PR #224, opened 2020-11-26
feat: add loading recordheader recordfield

amontalvof:


packages/record/src/components/RecordField/index.js
```diff
@@ -1,18 +1,42 @@
...
+                privateVariant={privateVariant}
+                type={type}
+                icon={icon}
+                iconPosition={iconPosition}
+            >
+                <RenderIf isTrue={isLoading}>
+                    <StyledLoadingValue />
+                </RenderIf>
+                <RenderIf isTrue={!isLoading}>
+                    <IconContainer iconPosition={iconPosition}>{icon}</IconContainer>
```
LeandroTorresSicilia, line 37: what happens when no icon is passed?

packages/record/src/components/RecordHeader/index.js
```diff
@@ -12,21 +13,43 @@ import {
...
             <Header>
                 <Details>
                     <RowContainer>
-                        <IconContainer>{icon}</IconContainer>
-                        <Label>{label}</Label>
+                        <RenderIf isTrue={isLoading}>
+                            <LabelLoadingShape />
+                        </RenderIf>
+                        <RenderIf isTrue={!isLoading}>
+                            <IconContainer>{icon}</IconContainer>
```
LeandroTorresSicilia, line 41: same ^^ here we are creating a dom element when nothing will render if an icon is not passed

LeandroTorresSicilia (approved):

Issue #223, opened 2020-11-22
docs: improve examples on `@rainbow-modules/record`


TahimiLeonBravo:
#### Designs
https://zpl.io/aX004ex
https://zpl.io/a7rrvvM
https://zpl.io/adZZko7
https://zpl.io/VY36XqM

PR #226, opened 2020-12-02
feat: added types in recordfield

amontalvof:
fix: #223


packages/record/src/components/RecordField/helpers/valueFormater.js
```diff
@@ -0,0 +1,40 @@
+export const currency = (value, currency) => {
+    return new Intl.NumberFormat(undefined, { style: 'currency', currency }).format(value);
+};
+
```
LeandroTorresSicilia, line 4: it is formatter and we should call this file valueFormatters

packages/record/src/components/RecordField/value.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { RenderIf } from 'react-rainbow-components';
+import * as formaterValueMap from './helpers/valueFormater';
+
+const Value = (props) => {
+    const { type, value, currency, href, onClick } = props;
+    const isUrl = type === 'url';
+    const formaterValue = formaterValueMap[type] || formaterValueMap.text;
```
LeandroTorresSicilia, line 9: formatter

packages/record/src/components/RecordField/value.js
```diff
@@ -0,0 +1,49 @@
...
+    const formatedValue = formaterValue(value, currency);
```
LeandroTorresSicilia, line 10: formattedValue

packages/record/docs/stories/RecordDetails.story.js
```diff
@@ -16,7 +16,7 @@ const Container = styled.div`
 `;
 
 const useChangeLoading = () => {
-    const [loading, setLoading] = useState(true);
+    const [loading, setLoading] = useState(false);
```
LeandroTorresSicilia, line 19: if you change this to false then when it is true?

LeandroTorresSicilia (approved):

Issue #235, opened 2020-12-15
feat: implement `CopyToClipboardButton` component on `@rainbow-modules/record` module


TahimiLeonBravo:
### Design
https://zpl.io/2jAQ8ex

### Usage
```
<CopyToClipboardButton value="foo" />
```

### Requirements
- [ ] it should have the keys to translate the tooltip text

PR #246, opened 2020-12-17
feat: implement `CopyToClipboardButton` component on `@rainbow-modules/record` module

yvmunayev:
#235


packages/record/src/components/CopyToClipboardButton/index.js
```diff
@@ -0,0 +1,152 @@
...
+    );
+}
+
+CopyToClipboardButton.propTypes = {
+    /** Text to be copied to clipboard */
+    value: PropTypes.string,
+    /** Text for the tooltip that describes the action of copying to clipboard. */
+    copyText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** text for the tooltip that describes that the text was copied to the clipboard. */
+    copiedText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
LeandroTorresSicilia, line 102: these texts will be translated by keys

packages/record/src/components/CopyToClipboardButton/index.js
```diff
@@ -0,0 +1,152 @@
...
+        triggerElementRef: () => triggerRef.current.htmlElementRef,
+    });
+
+    if (!isSupported()) {
+        return null;
+    }
+
+    const handleClick = () => {
+        const success = copy(value);
+        triggerRef.current.focus();
```
LeandroTorresSicilia, line 52: why need to focus manually?

packages/record/src/components/CopyToClipboardButton/index.js
```diff
@@ -0,0 +1,152 @@
...
+        if (success) {
+            setCopied(true);
+            setVisible(true);
+        }
+    };
+
+    const hiddenTooltip = () => {
```
LeandroTorresSicilia, line 59: this is an action then we should call it `hideTooltip`

packages/record/src/components/CopyToClipboardButton/helpers/__test__/isSupported.spec.js
```diff
@@ -0,0 +1,21 @@
...
+describe('<isSupported/>', () => {
+    it('should return true when window.clipboardData and window.clipboardData.setData is defined', () => {
+        window.clipboardData = {
+            setData: jest.fn(),
+        };
+        expect(isSupported()).toBeTruthy();
+    });
+    it('should return true when document.queryCommandSupported is defined and document.queryCommandSupported("copy") return true', () => {
+        document.queryCommandSupported = jest.fn(() => true);
+        expect(isSupported()).toBeTruthy();
```
LeandroTorresSicilia, line 12: here window.clipboardData.setData does not exists by default?

yvmunayev: window.clipboardData.setData is an Internet Explore API

packages/record/src/components/CopyToClipboardButton/index.js
yvmunayev: In the copy function when doing this:

``` js
const range = document.createRange ();
const selection = document.getSelection ();

range.selectNodeContents (mark);
selection.addRange (range);
```

the text is selected to copy it, the focus moves to this element so it must be changed manually.

packages/record/src/components/CopyToClipboardButton/index.js
```diff
@@ -26,7 +25,9 @@ export default function CopyToClipboardButton(props) {
     const tooltipRef = useRef();
     const [isVisible, setVisible] = useState(false);
     const [isCopied, setCopied] = useState(false);
-    const text = isCopied ? copiedText : copyText;
+    const copyProps = { id: 'click_to_copy', defaultMessage: 'Click to copy' };
+    const copiedProps = { id: 'copied', defaultMessage: 'Copied' };
```
LeandroTorresSicilia, line 29: look at other components like `packages/payment/src/components/CreditCardPicker/messages.js`
where we define a messages file and then set an id following a format to and then document these ids in order to let the user use them to translate with their custom dictionarie

LeandroTorresSicilia (approved):

PR #247, opened 2020-12-18
fix: `RecordField` with `type="dateTime" crashes when pass a string date or a new Date(string)

yvmunayev:
#240

### DIFF
--- a/packages/record/src/components/RecordField/helpers/valueFormatter.js
+++ b/packages/record/src/components/RecordField/helpers/valueFormatter.js
@@ -1,4 +1,4 @@
-import { isNumber } from '@rainbow-modules/validation';
+import { isNull, isNumber } from '@rainbow-modules/validation';
 
 export const currency = (value, currency) => {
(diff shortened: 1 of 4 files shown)


packages/record/src/components/RecordField/helpers/valueFormatter.js
```diff
@@ -12,27 +14,43 @@ export const percent = (value) => {
...
+    if (isValidDate(date)) {
+        return new Intl.DateTimeFormat(undefined, {
+            year: 'numeric',
+            month: 'numeric',
+            day: 'numeric',
+            hour: 'numeric',
+            minute: 'numeric',
+            second: 'numeric',
+        }).format(date);
+    }
```
LeandroTorresSicilia, line 32: here we should add a simple test mounting a the RecordField with ```type="dateTime"``` and passing an invalid date like a string and checks that it renders an empty string and does not crash

LeandroTorresSicilia (approved):

PR #256, opened 2020-12-23
feat: poc for new bar chart

HellWolf93:
This is the POC for the new chart. Suggestions are welcome.

The idea is to build a chart component with more flexibility in the configuration, while avoiding to have to manage a complex options object for Chart.js.

This is a work in progress, keep in mind:
- There is a lot of config options in Chart.js still missing, until now I've ported only the necessary to fulfill the example
- Docs for the prop-types are almost 100% missing
- We need to add a few more stories
- And we need a lot of tests

I will keep working on this, but early feedback is much appreciated ;)


LeandroTorresSicilia:
we gonna close this for now

PR #282, opened 2021-01-04
feat: cypress tests in globalsearch

amontalvof:


cypress/integration/search/globalSearch.js
```diff
@@ -0,0 +1,38 @@
+const GLOBAL_SEARCH_URL =
+    '/iframe.html?id=modules-search-stories-globalsearch--algolia-global-search';
+const AGOLIA_URL = 'https://6mt6rezjme-dsn.algolia.net/1/indexes/rainbow-books/query';
+const SEARCH_INPUT = 'input[type="search"]';
+
+const searchBooksAlgolia = () =>
```
LeandroTorresSicilia, line 6: I would call this algoliaSearchMock

cypress/integration/search/globalSearch.js
```diff
@@ -0,0 +1,38 @@
...
+    });
+
+    it('should open the GlobaSearch', () => {
+        cy.get(SEARCH_INPUT).click();
+        cy.get('h1').contains('Type something to search').should('exist');
+    });
+
+    it('should close the GlobaSearch', () => {
+        cy.get(SEARCH_INPUT).click();
+        cy.get('button').click();
```
LeandroTorresSicilia, line 24: this is the close button? If so, we should call the test `should close the GlobaSearch when clicking the close button`, also we should find the button in a more deterministic way, we can discuss it

LeandroTorresSicilia (approved):

PR #302, opened 2021-01-20
test: for `UniversalFormModal` component

yvmunayev:
fix: #287


packages/hooks/src/modal/__test__/useOpenModal.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+describe('useOpenModal', () => {
+    beforeAll(() => {
+        setProps.mockClear();
+    });
+    it('should not call setProps when the store does not have defined the key', () => {
+        const [openModal, closeModal] = useOpenModal('test');
+        openModal();
+        closeModal();
+        expect(setProps).toBeCalledTimes(0);
```
LeandroTorresSicilia, line 18: check this, since it can be a false positive since you are not setting setProps, then it will never be called

yvmunayev: @LeandroTorresSicilia what I am testing is that if you call useOpenModal ('test') without having previously called useConnectModal ('test') you should not call setProps

``` js
const openModal = (props) => {
        if (store.has(modalId)) {
            const [defaultProps, setProps] = store.get(modalId);
            setProps({
                ...initialDefaultProps,
                ...defaultProps,
                ...props,
                isOpen: true,
                onRequestClose: () =>
                    setProps({ ...initialDefaultProps, ...defaultProps, isOpen: false }),
            });
        }
    };
```

LeandroTorresSicilia (approved):

Issue #298, opened 2021-01-18
feat: implement editable on `ColoredStatusColumn` component


TahimiLeonBravo:
### Requirements
- [x] implement functionality
- [x] add story
- [ ] add test

#### Design
https://zpl.io/Vx7ny83

#### Specs
- [x] It should show a right arrow when `isEditable` prop is passed to the `Column` component
- [x] It should make the cell clickable when the `isEditable` prop is passed to the `Column` component
- [x] It should show a dropdown with the options when clicking on the cell

PR #305, opened 2021-01-22
feat: implement editable on `ColoredStatusColumn` component

HellWolf93:
fix: #298


packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -1,30 +1,95 @@
...
+            <RenderIf isTrue={isEditable}>
+                <StyledIndicator color={color} />
+                <InternalOverlay
+                    isVisible={isOpen}
+                    positionResolver={positionResolver}
+                    triggerElementRef={() => containerRef}
+                >
+                    <StyledDropdown value={{ name: value }} onChange={handleChange}>
+                        <HeaderOption />
+                        {getOptions()}
```
LeandroTorresSicilia, line 85: let's use this declarative:
```<Options />```
create Options components that make the iteration

File: packages/listview/src/components/ColoredStatusColumn/index.js [hunk 2]
```diff
@@ -1,30 +1,95 @@
...
+const ColoredStatusColumn = ({ row, value, colors, textTransform, isEditable, onChange }) => {
+    const [isOpen, setIsOpen] = useState(false);
+    const containerRef = useRef();
 
-const ColoredStatusColumn = ({ value, colors, textTransform }) => {
     const defaultColors = useDefaultColors();
     const map = colors ? getNormalizedColors(colors) : defaultColors;
     const backgroundColor = getBackgroundColor(map[value.toLowerCase()]);
     const color = getColor(map[value.toLowerCase()]);
+    const hasPopup = isEditable ? 'listbox' : null;
```
LeandroTorresSicilia, line 22: since this is not a boolean I think we should use other name: ariaHaspopup

packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -1,30 +1,95 @@
...
+            const color = getColor(map[value.toLowerCase()]);
+            const backgroundColor = getBackgroundColor(map[value.toLowerCase()]);
+            return (
+                <Option
+                    name={value}
+                    label={value}
+                    color={color}
+                    backgroundColor={backgroundColor}
+                    textTransform={textTransform}
+                    component={ColoredOption}
```
LeandroTorresSicilia, line 59: pass a unique key to the mapping items

packages/listview/src/components/ColoredStatusColumn/option.js
```diff
@@ -0,0 +1,52 @@
...
+    label: undefined,
+    color: undefined,
+    backgroundColor: undefined,
+    textTransform: 'capitalize',
+    isActive: false,
+    isSelected: false,
+};
+
+const HeaderOption = () => {
+    return <StyledHeader>Change Status</StyledHeader>;
```
TahimiLeonBravo, line 49: We should add a new prop for the header because it should support internationalization

TahimiLeonBravo (changes requested):

TahimiLeonBravo:
@HellWolf93 we should return the focus to the cell when close the dropdown

TahimiLeonBravo:
@HellWolf93 we should add some styles to the cell selected

TahimiLeonBravo:
@HellWolf93 we should remove the gray color outside the status option
Screen Shot 2021-01-22 at 11 48 08 PM

packages/listview/src/components/ColoredStatusColumn/index.d.ts
```diff
@@ -1,15 +1,32 @@
...
-    colors?: Record<string, string> | ColorObject;
+    colors?: ColorType;
     /** A string that indicates the format of the font capitalize | uppercase | lowercase */
     textTransform?: string;
+    /** A boolean that specifies whether a column is editable or not. Its default value is false.  */
+    isEditable?: boolean;
+    /** The action triggered when a value attribute changes. */
+    onChange?: (value: ChangeValue) => void;
+    /** Text or component to show ate the top of the dropdown */
+    dropdownHeader: ReactNode;
```
LeandroTorresSicilia, line 29: this is not required

packages/listview/src/components/ColoredStatusColumn/index.js
```diff
@@ -1,30 +1,98 @@
...
+    const containerRef = useRef();
+    const dropdownRef = useRef();
 
-const ColoredStatusColumn = ({ value, colors, textTransform }) => {
     const defaultColors = useDefaultColors();
     const map = colors ? getNormalizedColors(colors) : defaultColors;
     const backgroundColor = getBackgroundColor(map[value.toLowerCase()]);
     const color = getColor(map[value.toLowerCase()]);
+    const ariaHaspopup = isEditable ? 'listbox' : null;
+    const header = dropdownHeader || dropdownHeaderLabel;
```
LeandroTorresSicilia, line 34: we need to discuss this prop name and then if we pass this as a prop then we don't need the dropdownHeaderLabel

Issue #301, opened 2021-01-20
feat: implement `filterByFields` function


TahimiLeonBravo:
### Requierements

- Implement `filterByFields` function on `rainbow-modules@listview`

### Usage

```
filterByFields({

    data: [{ label: 'some string', obj: {. foo: {. bar: 'other string' } } }, ...],

.   query: 'string',

    fields: ['label', 'obj.foo.bar', ...]

});

```

### Specs

- `data`:  is the data collection where you want to search
- `query`: is the value that is being typing
- `fields`: the data fields where you want to search

﻿

PR #313, opened 2021-01-26
feat: implement search funcionality listview

amontalvof:
fix: #301


packages/listview/src/helpers/filterByFields.js
```diff
@@ -0,0 +1,47 @@
...
+        })
+        .join(',');
+};
+
+const filterByFields = (params) => {
+    const { query, data, fields } = params;
+    if (query) {
+        return data.filter((item) => {
+            const stringToMatch = mapValuesToString(item, fields);
+            const words = getWords(query);
```
yvmunayev, line 37: you can move getWords out of the filter

packages/listview/docs/stories/FilterByFields.story.js
```diff
@@ -0,0 +1,115 @@
...
+
+const Title = styled.h1`
+    font-size: 30px;
+`;
+
+export const FilterByFieldsExample = () => {
+    const [value, setValue] = useState('');
+    const [data, setData] = useState(initialData);
+
+    const handleOnChange = (value) => {
```
yvmunayev, line 79: handleChange

packages/listview/src/helpers/filterByFields.js
```diff
@@ -0,0 +1,47 @@
...
+
+const mapValuesToString = (obj, fields) => {
+    return fields
+        .map((field) => {
+            return getFieldValue(obj, field);
+        })
+        .join(',');
+};
+
+const filterByFields = (params) => {
```
LeandroTorresSicilia, line 32: now we are adding typescript declaration files, same as we do in react-rainbow

LeandroTorresSicilia:
create the documentation for this helper

LeandroTorresSicilia:
fix pr name to use conventional commit since commit-lint-pr is failing

yvmunayev (approved):

packages/listview/docs/stories/FilterByFields.story.js
```diff
@@ -0,0 +1,115 @@
...
+    const [data, setData] = useState(initialData);
+
+    const handleChange = (value) => {
+        setValue(value);
+        const filteredData = filterByFields({
+            data: initialData,
+            query: value,
+            fields: ['customer', 'email', 'trackingNumber', 'obj.status'],
+        });
+        setData(filteredData);
```
LeandroTorresSicilia, line 86: here you can use only one state (value), since data can be computed, if you want to avoid filterByFields each time because can be expensive then, you can use useMemo passing in the dependency array the value, thus only filterByFields will be called when the value change

LeandroTorresSicilia (approved):

PR #316, opened 2021-01-27
feat: add FieldsGenerator component to form module

LeandroTorresSicilia:


packages/forms/docs/stories/FieldsGenerator.story.js
```diff
@@ -0,0 +1,98 @@
...
+        errors.email = 'Looks like you forget to add your email';
+    }
+
+    return errors;
+};
+
+export const Basic = () => {
+    return (
+        <Application>
+            <FinalForm onSubmit={(values) => alert(JSON.stringify(values))} validate={validate}>
```
yvmunayev, line 82: @LeandroTorresSicilia why don't you use UniversalForm here

LeandroTorresSicilia: you are right, I will update it

Issue #314, opened 2021-01-26
fix: improve interaction on `FloatingSearchButtonIcon` component


TahimiLeonBravo:
#### Specs
- [ ] Implement functionality
- [ ] add test

#### Design
- https://zpl.io/V0KR0l9
- https://zpl.io/bArZE9B
- https://zpl.io/brD8mWO
- https://zpl.io/aN64dE4

#### Requirements
- [x] It should show only the icon when the search is empty and is not focus
- [x] it should show the "Filter" text when the floating input is visible
- [x] it should show the "Circle" icon when is not empty and is closed
- [x] it should close the floating input if is open and clicking outside
- [x] it should close the floating input if is empty and press the "x" button on the input
- [x] it should close de floating input if is open and click on the trigger button
- [x] it should close the floating input if is empty and press the "Esc" key
- [x] as a developer I should be able to customize the "Icon"
- [x] as a developer I should be able to change the "Filter" text

PR #317, opened 2021-01-29
fix: improve interaction on `FloatingSearchButtonIcon` component

wildergd:
fix: #314 

####Changes:
 - add close when ESC key is pressed
 - close when click outside
 - add `label` prop


packages/listview/src/components/FloatingSearch/index.js
```diff
@@ -14,7 +30,12 @@ const FloatingSearch = (props) => {
             onOpened={() => inputRef.current.focus()}
             keepScrollEnabled
         >
-            <Search inputRef={inputRef} {...rest} />
+            <Search
+                ref={searchBoxRef}
+                inputRef={inputRef}
+                {...rest}
```
LeandroTorresSicilia, line 36: here we should put ```{...rest}``` at top thus avoiding someone pass inputRef and cancel the prop

LeandroTorresSicilia:
some changes:
- we should use an icon by default
-  add filter text when the search is open  (add a key to internacionalization)
- when the search is open and types something then when press ESC we should clear the input, if the search does not have value then ESC will close the search

packages/listview/src/components/FloatingSearch/useKeyHandler.js
```diff
@@ -0,0 +1,24 @@
...
+export default function useKeyHandler(props) {
+    const { onRequestClose } = props;
+    const keyHandlerMap = useMemo(
+        () => ({
+            [ESCAPE_KEY]: () => onRequestClose(),
+        }),
+        [onRequestClose],
+    );
+
+    const keyDownHandler = (event, childProps) => {
```
LeandroTorresSicilia, line 13: what are childProps used for?

packages/listview/src/components/FloatingSearchButtonIcon/styled.js
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components';
+import { CircleFilled } from '@rainbow-modules/icons';
```
LeandroTorresSicilia, line 2: here instead of add an icon for this, use a span with a background and border-radius to achieve this

LeandroTorresSicilia (approved):

PR #323, opened 2021-02-02
fix: force disable legend in `DoubleBarChart` component

wildergd:


packages/dashboard/src/components/DoubleBarChart/helpers/getOptions.js
```diff
@@ -2,6 +2,11 @@ export default function getOptions(options) {
     const { xLabel, yLabel, suggestedMax } = options;
     return {
         maintainAspectRatio: false,
+        legend: {
+            display: true,
+            position: 'top',
+            align: 'start',
+        },
```
LeandroTorresSicilia, line 9: let's leave this legend as before, showing it in the bottom

LeandroTorresSicilia (approved):

Issue #319, opened 2021-01-30
feat: implement `UniversalFormOverlay` component on `@rainbow-modules/form`


TahimiLeonBravo:
### Design
https://zpl.io/2pwn6Z8

### Requirements:
- It should use the InternalOverlay component
- It should have a custom position resolver since this should always cover the element that triggers it.

### API
- `triggerElementRef`: Ref or function that returns a ref to a DOM element, the DOM element resolved by this ref will be used to positioning the component passed when visible.
One of type: `object` | `func`
- `isOpen`:  When true the Modal opens. `bool`
- `fields`: Component Class or Function with the fields of your form. Use Field component of react-final-form. `func`
- `onSubmit`: Callback that will be invoked after on submit event, it will receive an object with all the values of the form already validated. `func`
- `onRequestClose`: Callback that will be invoke when close event. Click on Cancel or Close Button or press ESC. `func`
- `initialValues`: The initial values of the form e.g. { name: 'Max', age: 30 }  `object`
- `submitButtonLabel`: The label of the submit button on the form. 
one of type: `string` | `node`
- `cancelButtonLabel` :  The label of the cancel button on the form. 
one of type: `string` | `node`
- `onOpened`: A callback triggered when the overlay component is opened. This is useful for example to set focus to an element after it is opened.

PR #325, opened 2021-02-04
feat: implement UniversalFormOverlay component

HellWolf93:
fix: #319


packages/forms/docs/stories/UniversalFormOverlay.story.js
```diff
@@ -0,0 +1,72 @@
...
+
+const Label = styled.span`
+    font-size: 1rem;
+    font-weight: bold;
+`;
+
+const Fields = () => {
+    return (
+        <div>
+            <Label>Name</Label>
```
LeandroTorresSicilia, line 32: I think we should use the label prop of the input, check with @TahimiLeonBravo if the styles need to be customized, then you can pass a node to the label prop

packages/forms/src/components/UniversalFormOverlay/__test__/universalFormOverlay.spec.js
```diff
@@ -0,0 +1,134 @@
...
+                    fields={Fields}
+                    cancelButtonLabel="Discard"
+                    isOpen
+                />
+            </Application>,
+        );
+        expect(wrapper.find('Button').first().text()).toBe('Discard');
+    });
+
+    it('should render submitButtonLabel inside cancel button', () => {
```
LeandroTorresSicilia, line 73: ... inside submit button

packages/forms/src/components/UniversalFormOverlay/index.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { useRef } from 'react';
+import { Button } from 'react-rainbow-components';
+import InternalOverlay from 'react-rainbow-components/components/InternalOverlay';
+import useOutsideClick from '../../../../hooks/src/useOutsideClick';
+import useUniqueIdentifier from '../../../../hooks/src/useUniqueIdentifier';
```
LeandroTorresSicilia, line 5: here use import as ```import { useOutsideClick, useUniqueIdentifier } from '@rainbow-modules/hooks';``` and then add `@rainbow-modules/hooks` as dependencies of this module

LeandroTorresSicilia (approved):

PR #326, opened 2021-02-08
feat: add isEditable prop to RecordField

HellWolf93:
fix: #320


TahimiLeonBravo:
@HellWolf93 let's add a `dirty` state to the component, it should look like this https://zpl.io/29p08vl

TahimiLeonBravo:
@HellWolf93 we should change the pencil color when hover the `RecordField`
https://zpl.io/2yKqkKw

packages/forms/src/components/UniversalFormOverlay/index.tsx
```diff
@@ -33,9 +39,24 @@ const UniversalFormOverlay: React.FC<UniversalFormOverlayProps> = ({
     };
 
     const handleKeyPress = (event: React.KeyboardEvent): void => {
-        if (event.key === 'Escape' && onRequestClose) {
+        if (isOpen && event.key === 'Escape' && onRequestClose) {
             onRequestClose();
         }
+        if (event.key === 'Tab' && containerRef.current !== undefined) {
+            manageTab(containerRef.current, event);
```
LeandroTorresSicilia, line 46: Great!! It is important this component has this feature

LeandroTorresSicilia (approved):

Issue #336, opened 2021-02-22
feat: implement `NotificationMenu` component on `rainbow-modules/notifications`


TahimiLeonBravo:
### Requirements
- [ ] Implement component
- [ ] add documentation
- [ ] add test

### Design
https://zpl.io/aR6Axoe

### API

```
<NotificationMenu
        label= string | node, (default="Notifications")
        footer= node
        onClick= function
        isLoading = boolean
        unreads= number | true | false
        notification = {
              title: 'string',
              description: 'string',
              status: 'success' | 'error' | 'warning' | 'inProgress' | node,
              createdAt: 'string',
              icon: node | string,
              unread
      }
 />

```

PR #340, opened 2021-02-26
feat: implement `NotificationMenu` component

HellWolf93:
fix: #336


packages/notifications/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "version": "0.29.0",
+    "main": "lib/index.js",
+    "author": "Nexxtway Team",
+    "license": "MIT",
+    "files": [
+        "lib"
+    ],
+    "dependencies": {
+        "styled-components": "^4.3.2",
+        "react-intl": "^4.4.0",
```
LeandroTorresSicilia, line 12: use react-intl as a peer dependency

packages/notifications/src/components/NotificationMenu/icons/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+import SuccessIcon from './successIcon';
+import WarningIcon from './warningIcon';
+import { IconProps } from '../types';
+import { StyledSpinner } from '../styled';
+
+const iconMap: Record<string, () => ReactElement> = {
+    info: () => <InfoIcon />,
+    error: () => <ErrorIcon />,
+    warning: () => <WarningIcon />,
+    success: () => <SuccessIcon />,
```
LeandroTorresSicilia, line 14: here you can use the icon declaration directly:
```success: SuccessIcon```

File: packages/notifications/docs/stories/NotificationMenu.story.js [hunk 3]
```diff
@@ -0,0 +1,114 @@
...
+        title: 'Rainbow Components',
+        description: 'Lorem ipsum dolor sit amer, adipiscing consectetur',
+        status: 'success',
+        createdAt: 'Now',
+    },
+    {
+        title: 'Rainbow Modules',
+        description: 'Lorem ipsum dolor sit amer, adipiscing consectetur',
+        status: 'success',
+        createdAt: '3 minutes ago',
```
LeandroTorresSicilia, line 18: here createdAt should be a date or timestamp, also since this will be connected with functions with firebase then we should support firebase timestamps

packages/notifications/src/components/NotificationMenu/notifications.tsx
```diff
@@ -0,0 +1,43 @@
...
+    if (isLoading) {
+        return (
+            <>
+                <Notification isLoading />
+                <Notification isLoading />
+            </>
+        );
+    }
+
+    const children = notifications ? (
```
LeandroTorresSicilia, line 20: here we should show the empty message when passing an empty array

PR #359, opened 2021-03-18
fix: close notificationMenu when a notification is clicked

wildergd:


packages/notifications/src/components/NotificationMenu/index.tsx
```diff
@@ -41,6 +41,13 @@ const NotificationMenu: React.FC<NotificationMenuProps> = ({
         setIsOpen(!isOpen);
     };
 
+    const handleItemClick = () => {
+        setIsOpen(!isOpen);
+        if (onClick) {
+            onClick();
```
LeandroTorresSicilia, line 47: check here since in the onClick we are sending data

LeandroTorresSicilia (approved):

PR #361, opened 2021-03-19
feat: expose close method via ref in NotificationMenu component

wildergd:


packages/notifications/src/components/NotificationMenu/index.tsx
```diff
@@ -1,86 +1,101 @@
...
+            label,
+            footer,
+            unreads,
+            isLoading,
+            notifications,
+            onClick,
+            className,
+            style,
+        }: NotificationMenuProps,
+        ref,
```
LeandroTorresSicilia, line 31: in order to make this more readable you can pass here props and ref, then make props destructuring inside the function body

LeandroTorresSicilia (approved):

PR #366, opened 2021-03-24
fix: improved use unique identifier hook

amontalvof:


packages/hooks/src/useUniqueIdentifier.js
```diff
@@ -1,9 +1,8 @@
 import { useMemo } from 'react';
+import { nanoid } from 'nanoid';
 
-let idCounter = 0;
 export default function useUniqueIdentifier(prefix) {
     return useMemo(() => {
-        idCounter += 1;
-        return prefix ? `${prefix}-${idCounter}` : String(idCounter);
+        return `${prefix}-${nanoid(5)}`;
```
LeandroTorresSicilia, line 6: here you should check if prefix is passed to not attach the `-` if not passed

LeandroTorresSicilia (approved):

LeandroTorresSicilia (approved):

LeandroTorresSicilia (approved):

PR #377, opened 2021-05-07
feat: add message support to app spinner

wildergd:


packages/app/src/actions/index.d.ts
```diff
@@ -1,6 +1,9 @@
 import { ReactNode } from 'react';
 
-export function showAppSpinner(): void;
+type AppMessagepParams = {
+    message: string;
```
LeandroTorresSicilia, line 4: it should be optional

LeandroTorresSicilia, line 4: also use ReactNode

LeandroTorresSicilia (approved):

PR #386, opened 2021-05-27
feat: implement PageRating component

HellWolf93:


packages/content/src/components/PageRating/option.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { StyledInput, StyledLabel } from './styled';
+import { IconProps } from './types';
+
+const Option: React.FC<IconProps> = ({ isSelected, value, onChange, children }: IconProps) => {
+    return (
+        <StyledLabel isSelected={isSelected}>
+            <StyledInput
+                type="radio"
+                name="pageRating"
```
LeandroTorresSicilia, line 10: we used to pass a name prop to use here and if it is not passed then we generate a one with our uniqueId helper, in this case you should use ```useUniqueIdentifier```

packages/content/src/components/PageRating/styled.ts
```diff
@@ -0,0 +1,90 @@
+import styled from 'styled-components';
+import { LabelAlignments } from './types';
+
+export const StyledContainer = styled.div<{ labelAlignment?: LabelAlignments }>`
```
LeandroTorresSicilia, line 4: the container should be a `fieldset` element

packages/content/src/components/PageRating/styled.ts
```diff
@@ -0,0 +1,90 @@
...
+        flex-direction: row-reverse;
+        align-items: center;
+        > h3 {
+            margin-bottom: 0;
+            margin-left: 14px;
+        }
+    `}
+`;
+
+export const StyledLegend = styled.h3`
```
LeandroTorresSicilia, line 34: here use `legend` element

packages/content/src/components/PageRating/styled.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const StyledLabel = styled.label<{ isSelected?: boolean }>`
+    position: relative;
+    display: inline-flex;
+    margin: 0 4px;
+    cursor: pointer;
+    transition-property: transform;
+    transition-duration: 300ms;
+    transform-origin: center;
+    color: #d3dadf;
```
LeandroTorresSicilia, line 63: @TahimiLeonBravo check this inline color

packages/content/src/components/PageRating/styled.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    :last-child {
+        margin-right: 0;
+    }
+
+    ${(props) =>
+        props.isSelected &&
+        `
+    color: #ffca28;
```
LeandroTorresSicilia, line 82: @TahimiLeonBravo same ^^

packages/content/src/components/PageRating/types.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface IconProps {
+    value?: string | number;
+    isSelected?: boolean;
+    onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
+    children?: ReactNode;
+}
+
+export interface PageRatingProps {
+    className?: string;
+    style?: Record<string, unknown>;
```
LeandroTorresSicilia, line 16: there a CSSProperties interface used for style that is imported from react

PR #390, opened 2021-06-02
feat: implement CodeBlock component

HellWolf93:


packages/content/src/components/CodeBlock/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+            </StyledContent>
+        </StyledContainer>
+    );
+};
+
+CodeBlock.defaultProps = {
+    label: undefined,
+    value: undefined,
+    language: undefined,
+    showLineNumbers: true,
```
LeandroTorresSicilia, line 39: let's use `false` by default
normally we use to set `false` as default when the prop is `showSomething`, but when we want to show it by default then the prop should be name as `hideSomething` with a default value of `false`, this avoiding to use something like:
```
showSomething={false}
```
which is hard to understand

LeandroTorresSicilia (approved):

PR #392, opened 2021-06-08
feat: implement ResizableColumns component

HellWolf93:


TahimiLeonBravo (approved):

packages/layout/src/components/ResizableColumns/types.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export interface ResizableColumnsProps {
+    /**
+     * Sets the initial width for the main column.
+     *
+     * By default the main column is the left one. Pass a value less than zero to set the right column
+     * as the main one.
+     */
+    initialDividerWidth?: number;
```
LeandroTorresSicilia, line 35: we talk about this and we will use `initialDividerPosition`, then in the description explain that when passing a positive number then it is the divider position respect to the left, and when the number is less than zero then it is the divider position respect to the right

packages/layout/src/components/ResizableColumns/types.ts
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * The minimum width of the right column.
+     */
+    minRightWidth?: Width;
+    /**
+     * Function to invoke when the user finishes resizing.
+     *
+     * Useful to store the size of the columns.
+     */
+    onResize?: ({ dividerWidth }: OnResizeParams) => void;
```
LeandroTorresSicilia, line 61: ```{ dividerPosition }```

packages/layout/src/components/ResizableColumns/types.ts
```diff
@@ -0,0 +1,62 @@
...
+    initialDividerWidth?: number;
+}
+
+export type Width = {
+    px?: number;
+    percent?: number;
+};
+
+type OnResizeParams = {
+    dividerWidth?: number;
```
LeandroTorresSicilia, line 25: dividerPosition

PR #395, opened 2021-06-14
feat: implement StackedAccordion component

HellWolf93:


TahimiLeonBravo (approved):

packages/layout/src/components/StackedAccordionSection/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+    StyledSectionContent,
+    StyledSectionHeader,
+    StyledResizeBar,
+} from './styled';
+import { StackedAccordionSectionProps, MoveProps } from './types';
+import context from '../StackedAccordion/context';
+import { Context } from '../StackedAccordion/types';
+import RightArrow from './rightArrow';
+import { useResize } from './hooks/useResize';
+import shouldMove from '../ResizableColumns/helpers/shouldMove';
```
LeandroTorresSicilia, line 15: if this helper will be reusable by two o more components, then I think we should move it to a helper folder inside `src`

PR #400, opened 2021-06-17
feat: implement `useUserClaims` hook

HellWolf93:


packages/firebase-hooks/package.json
```diff
@@ -14,6 +14,7 @@
         "access": "public"
     },
     "dependencies": {
-        "react-query": "^3.5.5"
+        "react-query": "^3.5.5",
+        "firebase": "^8.2.0"
```
LeandroTorresSicilia, line 18: can we add this as peer dep?

LeandroTorresSicilia (approved):

PR #403, opened 2021-06-20
fix: improve `CodeBlock` component

HellWolf93:


TahimiLeonBravo:
@HellWolf93 also add the `style`, `className` prop

packages/icons/src/index.d.ts
```diff
@@ -103,3 +103,4 @@ export { default as Docs } from './components/docs';
 export { default as TwitterCircle } from './components/twitterCircle';
 export { default as Microsoft } from './components/microsoft';
 export { default as Yahoo } from './components/yahoo';
+export { default as Copy } from './components/copy';
```
LeandroTorresSicilia, line 106: we need to add this icon to de filter array in order to be searchable in the storybook UI

LeandroTorresSicilia (approved):

TahimiLeonBravo (approved):

PR #404, opened 2021-06-24
feat: implement `LinkButton` component

HellWolf93:


packages/misc/src/components/LinkButton/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+LinkButton.defaultProps = {
+    className: undefined,
+    style: undefined,
+    children: null,
+    variant: 'neutral',
+    shaded: false,
+    size: 'medium',
+    to: undefined,
+    target: undefined,
+    replace: false,
```
LeandroTorresSicilia, line 25: what is replace used for?

packages/misc/src/components/LinkButton/types.ts
```diff
@@ -0,0 +1,13 @@
+import { CSSProperties } from 'react';
+import { ButtonIconVariant } from 'react-rainbow-components/components/types';
+import { LinkProps } from 'react-router-dom';
+
+export interface ButtonProps {
+    className: string;
+    style: CSSProperties;
```
LeandroTorresSicilia, line 7: all props are optional

packages/misc/src/components/LinkButton/index.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { StyledLink } from './styled';
+import { LinkButtonProps } from './types';
+
+const LinkButton: React.FC<LinkButtonProps> = (props: LinkButtonProps) => {
+    const { to, children } = props;
+
+    return (
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        <StyledLink href={to} {...props}>
```
LeandroTorresSicilia, line 10: I think we should avoid using the spread and then pass only our specific props

packages/misc/src/components/LinkButton/types.ts
```diff
@@ -0,0 +1,13 @@
...
+    variant?: ButtonIconVariant;
+    shaded?: boolean;
+    size?: 'small' | 'medium' | 'large';
+}
+
+export type LinkButtonProps = React.PropsWithChildren<ButtonProps & LinkProps>;
```
LeandroTorresSicilia, line 13: using LinkProps means that the component will accept all the Link props and we will only use `to` for now

packages/misc/src/components/LinkButton/styled.js
```diff
@@ -0,0 +1,380 @@
...
+    FONT_SIZE_HEADING_MEDIUM,
+} from 'react-rainbow-components/styles/fontSizes';
+import { Link } from 'react-router-dom';
+
+// eslint-disable-next-line import/prefer-default-export
+export const StyledLink = attachThemeAttrs(
+    styled(({ className, style, children, to, target }) => {
+        const As = target === '_blank' ? 'a' : Link;
+        return (
+            <As className={className} style={style} to={to} href={to} target={target}>
```
LeandroTorresSicilia, line 19: when using anchor with ```target="_blank"``` we must also pass ```rel="noopener noreferrer"``` for security reasons

TahimiLeonBravo (approved):

LeandroTorresSicilia (approved):

PR #405, opened 2021-06-26
feat: implement `InputSearch` component

HellWolf93:
TODO:
- [ ] Add Enter key icon


LeandroTorresSicilia:
when press ESC button the input should clear its value

packages/search/src/components/InputSearch/__test__/inputSearch.spec.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import { Application } from 'react-rainbow-components';
+import InputSearch from '..';
+
+describe('<Trailing />', () => {
```
LeandroTorresSicilia, line 6: <InputSearch />

packages/search/src/components/InputSearch/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const trailingRef = useRef<HTMLDivElement>(null);
+    const inputRef = useRef<InputComponent>(null);
+    const [paddingRight, setPaddingRight] = useState<number>();
+
+    useLayoutEffect(() => {
+        if (trailingRef.current) {
+            const { width } = trailingRef.current.getBoundingClientRect();
+            if (paddingRight !== width) setPaddingRight(width);
+        }
+    }, [paddingRight, value]);
```
LeandroTorresSicilia, line 27: I think we can compute in a functional way the padding needed based on the props (value and variant) and then avoid using a state for the padding

packages/search/src/components/InputSearch/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <StyledInput
+                name={name}
+                icon={<MagnifyingGlass />}
+                paddingRight={paddingRight}
+                value={value}
+                disabled={disabled}
+                readOnly={readOnly}
+                onChange={handleChange}
+                onKeyDown={handleKeyDown}
+                ref={inputRef}
```
TahimiLeonBravo, line 53: to have all the functionality of a native search input we should pass it `type ="search"`

TahimiLeonBravo (changes requested):

PR #408, opened 2021-06-30
fix: added placeholder prop to input search

amontalvof:


LeandroTorresSicilia:
@amontalvof conflicts here

PR #415, opened 2021-07-10
feat: implement `VirtualizedTable` component

HellWolf93:


packages/listview/src/components/VirtualizedTable/types.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export type CallbackFn = (args: CallbackFnArgs) => void;
+
+export interface DataObject {
+    initialData: Data;
+    push: (data: Data) => void;
+    unshift: (data: Data) => void;
+    update: (data: Data) => void;
+    insert: (data: Data, index: number) => void;
+    delete: (index: number, deleteCount: number) => void;
```
LeandroTorresSicilia, line 22: push, unshift are clear to me, then let's create one called set that will override all the data, then let's create:
- updateById({ id, data })
- updateByIndex({ index, data })
- deleteById({ id })
- deleteByIndex({ index })
 then change push and unshift to:
- push({ data })
- unshift({ data })
- set({ data })
and remove delete and insert
This is what I think we should have for now, the when we use it we can change or add others methods

LeandroTorresSicilia (approved):

PR #434, opened 2021-08-03
fix: put the example docs outside the pageObject file

HellWolf93:


packages/cypress/src/pageObjects/Lookup/examples.mdx
```diff
@@ -0,0 +1,30 @@
...
+-   ##### How to pick an option
+
+```js
+import { Lookup } from '@rainbow-modules/cypress/pageObjects';
+
+const lookup = new Lookup('#lookup-selector');
+lookup.input.click().then(($input) => {
+    const ariaControls = $input.attr('aria-controls');
+    const dropdown = new InternalDropdown(`#${ariaControls}`);
+    dropdown.options[0].click();
```
LeandroTorresSicilia, line 28: here the user should not know about InternalDropdown, we need to define another way to select the options

LeandroTorresSicilia: we need to do something like:
```
lookup.options[0].click()
```

packages/cypress/src/pageObjects/Picklist/examples.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+const picklist = new Picklist('#picklist-selector');
+picklist.click();
+// We need to get the `aria-controls` attribute of the input
+picklist.input.then(($input) => {
+    const ariaControls = $input.attr('aria-controls');
+    // Use the `aria-controls` value to create a new InternalDropdown page object
+    const dropdown = new InternalDropdown(`#${ariaControls}`);
+    // You can use the InternalDropdown page object here
+    dropdown.options[1].click();
```
LeandroTorresSicilia, line 23: same ^^

packages/cypress/src/pageObjects/index.ts
```diff
@@ -1,9 +1,13 @@
...
 export { default as Option } from './Option';
 export { default as PhoneInput } from './PhoneInput';
 export { default as Picklist } from './Picklist';
-export { default as UniversalPicker } from './UniversalPicker';
+export { default as Sidebar } from './Sidebar';
+export { default as SidebarItem } from './SidebarItem';
 export { default as Table } from './Table';
-export { default as ButtonMenu } from './ButtonMenu';
+export { default as UniversalPicker } from './UniversalPicker';
+export { default as UniversalPickerOption } from './UniversalPickerOption';
```
LeandroTorresSicilia, line 13: good!!!

LeandroTorresSicilia (approved):

PR #440, opened 2021-09-04
fix: improve Picklist page object

HellWolf93:


packages/cypress/src/pageObjects/InternalDropdown/examples.mdx
```diff
@@ -7,6 +7,17 @@ const dropdown = new InternalDropdown('#dropdown-selector');
...
 
+-   ##### How to wait for an option to exist
+
+```js
+import { InternalDropdown } from '@rainbow-modules/cypress/pageObjects';
+
+const dropdown = new InternalDropdown('#dropdown-selector');
+dropdown.waitForOption(1).then(() => {
+    dropdown.options[1].click();
+});
```
LeandroTorresSicilia, line 18: instead of adding a waitForOption method, can we make something like this?
```
const options =  await dropdown.options;
options[1].click();
```

packages/cypress/src/pageObjects/InternalDropdown/examples.mdx
```diff
@@ -7,6 +7,16 @@ const dropdown = new InternalDropdown('#dropdown-selector');
 dropdown.options[1].click();
 ```
 
+-   ##### How to wait for an option to exist
+
+```js
+import { InternalDropdown } from '@rainbow-modules/cypress/pageObjects';
+
+const dropdown = new InternalDropdown('#dropdown-selector');
+const options = await dropdown.options;
```
LeandroTorresSicilia, line 16: if alway we will need to wait then update all other doc to uses this

packages/cypress/src/pageObjects/Picklist/examples.mdx
```diff
@@ -17,3 +17,13 @@ picklist.input.click().then(() => {
     picklist.options[0].click();
 });
```
LeandroTorresSicilia, line 18: update this doc

LeandroTorresSicilia (approved):

PR #441, opened 2021-09-05
feat: implement AppMessage page object

HellWolf93:


packages/cypress/src/pageObjects/AppMessage/index.ts
```diff
@@ -0,0 +1,71 @@
+/// <reference types="cypress" />
```
LeandroTorresSicilia, line 1: since AppMessage is a component inside the RainbowFirebaseApp, we should make a page object for RainbowFirebaseApp

packages/cypress/src/pageObjects/AppMessage/index.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.rootElement = rootElement;
+    }
+
+    /**
+     * Asserts if the message passed is the current app message
+     * @method
+     * @param {string} error - The message text, that must match the current app message
+     * @returns {Cypress.Chainable<JQuery<HTMLElement>>}
+     */
+    shouldHaveMessage(message: string): Cypress.Chainable<JQuery<HTMLElement>> {
```
LeandroTorresSicilia, line 33: we will change the way of doing assertions inside the page object and instead the page object will have an `expect` method
```
const app = new RainbowFirebaseApp();
app.message.expect('variant', 'error');
app.message.expect('message', 'some text');
app.message.expect('isVisible', true);
```

packages/cypress/src/pageObjects/AppMessage/index.ts
```diff
@@ -0,0 +1,71 @@
...
+     * Get the close button of the app message
+     * @method
+     * @returns {Cypress.Chainable<JQuery<HTMLButtonElement>>}
+     */
+    get closeButton(): Cypress.Chainable<JQuery<HTMLButtonElement>> {
+        return cy
+            .get(this.rootElement)
+            .find<HTMLButtonElement>('button[data-id="button-icon-element"]')
+            .first();
+    }
```
LeandroTorresSicilia, line 68: I know we have been doing in this way and we will continue, but in cases like this where the user can do only one thing, in this case, close the message, we will change the approach and instead create a close method that close the message app

LeandroTorresSicilia (approved):

PR #454, opened 2021-09-18
feat: implement FloatingSearch page object

HellWolf93:


packages/listview/src/components/FloatingSearch/search.js
```diff
@@ -49,7 +55,12 @@ const Search = forwardRef((props, ref) => {
                     </StyledClearButton>
                     <StyledDivider />
                 </RenderIf>
-                <ButtonIcon size="small" icon={<Close />} onClick={onRequestClose} />
+                <ButtonIcon
+                    id="close-search-button"
```
LeandroTorresSicilia, line 59: I think our components should not use id internally, instead, you can use data-* like data-id because id must be unique in the document and then we can't know what id the user will use

LeandroTorresSicilia (approved):

PR #464, opened 2021-09-30
fix: add languageResolver prop to CodeViewer component

wildergd:


packages/content/src/components/CodeViewer/renderers/virtualizedRenderer.jsx
```diff
@@ -0,0 +1,39 @@
...
+            stylesheet,
+            style,
+            useInlineStyles,
+            key,
+        });
+};
+
+const virtualizedRenderer = ({ overscanRowCount = 10, rowHeight = 15 } = {}) => {
+    return ({ rows, stylesheet, useInlineStyles }) => (
+        <div style={{ height: '100%' }}>
```
LeandroTorresSicilia, line 22: no inline styles

PR #486, opened 2021-11-02
docs: add examples to field validators

HellWolf93:


packages/forms/docs/stories/validators/validators.story.js
```diff
@@ -0,0 +1,197 @@
...
+                    labelAlignment="left"
+                    validate={minLength(50)}
+                    className="rainbow-m-bottom_large"
+                />
+            </UniversalForm>
+            <StyledButton label="Submit" type="submit" form="minLength" variant="brand" />
+        </Container>
+    </RainbowFirebaseApp>
+);
+
```
LeandroTorresSicilia, line 189: here we should also add examples using composeValidators and using more than one validator like for example when you want to validate an email, first you want to check if it required and then you want to check if it is a valid email

LeandroTorresSicilia (approved):

PR #487, opened 2021-11-02
feat: add spinner and animation when iframe shows

HellWolf93:


packages/micro/src/components/Iframe/types.ts
```diff
@@ -11,4 +11,6 @@ export interface IframeProps {
     title?: string;
     variant?: Variant;
     onRequestClose?: () => void;
+    isLoading?: boolean;
+    onLoad?: () => void;
```
LeandroTorresSicilia, line 15: isLoading and onLoad are exposed props?

HellWolf93: No they are not. Fixed.

LeandroTorresSicilia (approved):

PR #521, opened 2022-01-30
feat: implement AuditLogs component

HellWolf93:


packages/icons/src/index.d.ts
```diff
@@ -80,14 +80,20 @@ export { default as OneFilled } from './components/oneFilled';
 export { default as List } from './components/list';
 export { default as Seat } from './components/seat';
 export { default as FrontCar } from './components/frontCar';
-export { default as Update } from './components/update';
+export { default as Function } from './components/function';
+export { default as Phone } from './components/phone';
+export { default as Envelope } from './components/envelope';
+export { default as InteractiveExamples } from './components/interactiveExamples';
+export { default as YoutubeIcon } from './components/youtubeIcon';
+export { default as MediumIcon } from './components/mediumIcon';
```
LeandroTorresSicilia, line 88: I think we should export this icons as Youtube and Medium without the Icon suffix

packages/listview/package.json
```diff
@@ -12,9 +12,13 @@
     },
     "dependencies": {
         "@rainbow-modules/app": "^0.48.0",
+        "@rainbow-modules/content": "^0.48.0",
         "@rainbow-modules/firebase-hooks": "^0.48.0",
+        "@rainbow-modules/hooks": "^0.48.0",
         "@rainbow-modules/icons": "^0.48.0",
         "@rainbow-modules/validation": "^0.48.0",
+        "firebase": "^8.2.0",
```
LeandroTorresSicilia, line 20: since whatever project that uses this need to have installed firebase for define their own firebase config and pass it to the RainbowFirebaseApp component, then we should put firebase as a peer dep

packages/listview/src/components/AuditLogs/filterDropdown/filters.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { LabelFilter as LabelFilterType, PicklistValue } from '../types';
+import context from '../context';
+import getLabelFilters from './helpers/getLabelFilters';
+
+const FilterOptions = ({ labels = [] }: { labels: string[] }) => (
+    <>
+        {labels.map((label) => (
+            <Option name={label} label={label} />
+        ))}
+    </>
```
LeandroTorresSicilia, line 14: here you can return the map directly, then remove the fragment

packages/listview/src/components/AuditLogs/summary.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+type LabelsProps = { labels: Record<string, string> };
+
+const SummaryLabels = ({ labels }: LabelsProps) => {
+    return (
+        <>
+            {Object.keys(labels).map((name) => (
+                <Badge label={labels[name]} size="small" />
+            ))}
+        </>
```
LeandroTorresSicilia, line 19: same ^^

packages/listview/src/components/AuditLogs/types.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { ReactNode, RefObject } from 'react';
+
+export interface AuditLogsProps {
+    collectionPath: string;
+    labels?: string[];
+    defaultFilter?: Record<string, any>;
```
LeandroTorresSicilia, line 7: here we need to be more specific about this type

LeandroTorresSicilia:
Screen Shot 2022-01-30 at 4 47 40 PM
Here we need to format the date to see a readable date.
Also remove the id field from the firestore document, if you are using the `useCollection` then you can remove the `flat: true` prop thus removing the id field from the data

LeandroTorresSicilia:
instead of filtering in each change in the `filter by label`, we should use an apply button (check with @TahimiLeonBravo how it will look)

LeandroTorresSicilia:
Screen Shot 2022-01-30 at 4 52 31 PM
The `filter by label` menu should have border-radius like other menus

LeandroTorresSicilia:
https://user-images.githubusercontent.com/24737097/151719486-a46a266b-83fc-4326-a039-828cf34b31f5.mov

if click filter name and select one, if click same label again it is removed and you can't add other again

packages/listview/src/components/AuditLogs/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    component={FirestoreDate}
+                />
+                <Column
+                    header="Summary"
+                    field="textPayload"
+                    headerAlignment="left"
+                    cellAlignment="left"
+                    component={Summary}
+                />
+            </StyledTable>
```
LeandroTorresSicilia, line 103: here we should use the `FirestoreTableWithCursors` component in order to have pagination

packages/listview/src/components/AuditLogs/filterDropdown/filters.tsx
HellWolf93: This is no possible to achieve due to current limitations of TypeScript and Rect typings.

See [here](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20356).

PR #522, opened 2022-01-30
feat: implement column header filter text

yvmunayev:


packages/listview/docs/stories/ColumnHeaderFilterText.story.js
```diff
@@ -0,0 +1,107 @@
...
+        const newData = [...filteredData];
+        const key = (value) => value[sortedBy];
+        const reverse = sortDirection === 'asc' ? 1 : -1;
+
+        return newData.sort((aItem, bItem) => {
+            const aValue = key(aItem);
+            const bValue = key(bItem);
+            return reverse * ((aValue > bValue) - (bValue > aValue));
+        });
+    }, [filteredData, sortDirection, sortedBy]);
```
LeandroTorresSicilia, line 64: we should add another example more simple without handling sorting since all this logic pollute the example

packages/listview/docs/stories/ColumnHeaderFilterText.story.js
```diff
@@ -0,0 +1,107 @@
...
+                    variant="listview"
+                >
+                    <Column header="Id" field="id" />
+                    <Column
+                        header="Name"
+                        field="name"
+                        sortable
+                        headerComponent={(props) => (
+                            <ColumnHeaderFilterText onFilter={setFilters} {...props} />
+                        )}
```
LeandroTorresSicilia, line 89: the component is intended to be used like:
```
<Column
    header="Name"
    field="name"
    sortable
    headerComponent={ColumnHeaderFilterText}
    onFilter={setFilters}
/>
```

PR #523, opened 2022-01-30
feat: download modal component

amontalvof:


cypress/integration/listview/downloadModal.spec.js
```diff
@@ -0,0 +1,50 @@
...
+            'contain',
+            'The entries matching will be downloaded. If you need over 10,000 entries consider exporting.',
+        );
+        cy.get('h2').should('have.length', 2);
+        cy.get('h2').eq(0).should('contain', 'Maximum entries');
+        cy.get('h2').eq(1).should('contain', 'Select format');
+        cy.get('[type="number"]').should('have.value', '500');
+        cy.get('[type="radio"]').eq(0).should('be.checked').and('have.value', 'json');
+        cy.get('[type="radio"]').eq(1).should('not.be.checked').and('have.value', 'csv');
+        cy.get('button').contains('Download').click();
```
LeandroTorresSicilia, line 25: we need to scope these selectors maybe something like:
cy.get('#modal-id').within(() => {
// then here use all these selectors
})
for this, you need to pass a new prop id that is passed to the internal Modal

cypress/integration/listview/downloadModal.spec.js
```diff
@@ -0,0 +1,50 @@
...
+        cy.get('h2').should('have.length', 2);
+        cy.get('h2').eq(0).should('contain', 'Maximum entries');
+        cy.get('h2').eq(1).should('contain', 'Select format');
+        cy.get('[type="number"]').should('have.value', '500');
+        cy.get('[type="radio"]').eq(0).should('be.checked').and('have.value', 'json');
+        cy.get('[type="radio"]')
+            .eq(1)
+            .should('not.be.checked')
+            .and('have.value', 'csv')
+            .check({ force: true });
```
LeandroTorresSicilia, line 46: this is too repetitive, all this is already tested in the previous test

packages/listview/docs/stories/DownloadModal.story.js
```diff
@@ -0,0 +1,57 @@
...
+import { ButtonIcon } from 'react-rainbow-components';
+import { RainbowFirebaseApp } from '@rainbow-modules/app';
+import { Download } from '@rainbow-modules/icons';
+import DownloadModal from '../../src/components/DownloadModal';
+import app from '../../../../firebase';
+
+export const DefaultDownloadModal = () => {
+    const [isOpen, setIsOpen] = useState(false);
+
+    const handleDownload = async (max, format) => {
```
LeandroTorresSicilia, line 11: max is not working here, if I select 3 as the max entries the example continues downloading all the data

packages/listview/docs/stories/DownloadModal.story.js
```diff
@@ -0,0 +1,57 @@
...
+            },
+            { name: 'Item 2', color: 'Red', size: 'X-Medium', max, format },
+            { name: 'Item 3', color: 'Blue', size: 'X-Small', max, format },
+            { name: 'Item 4', color: 'Orange', size: 'S-Medium', max, format },
+            { name: 'Item 5', color: 'Yellow', size: 'S-Small', max, format },
+            { name: 'Item 6', color: 'Cyan', size: 'S-Large', max, format },
+            { name: 'Item 7', color: 'Magenta', size: 'M-Small', max, format },
+            { name: 'Item 8', color: 'Violet', size: 'M-Large', max, format },
+            { name: 'Item 9', color: 'Brown', size: 'M-Medium', max, format },
+        ];
```
LeandroTorresSicilia, line 28: here we should simulate this as an async operation, for example:
```
const fetchData = () => {
    return new Promise((resolve) => {
        setTimeout(() => Promise.resolve(/* data here */), 2000);
    });
};
```

packages/listview/docs/stories/DownloadModal.story.js
```diff
@@ -0,0 +1,57 @@
...
+                id="download-button-icon"
+                variant="border"
+                icon={<Download />}
+                onClick={() => setIsOpen(true)}
+                tooltip="Download"
+            />
+            <DownloadModal
+                isOpen={isOpen}
+                onRequestClose={() => setIsOpen(false)}
+                maxEntries="500"
```
LeandroTorresSicilia, line 43: this should be a number, we can let the user pass a string but the right way and what we should use in this example is a number

packages/listview/src/components/DownloadModal/index.d.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable @typescript-eslint/ban-types */
+import { CSSProperties } from 'react';
+
+type Max = string | number;
+
+export interface DownloadModalProps {
+    /** The title at the top of the DownloadModal component. */
+    title?: string;
+    /** The description in the DownloadModal component. */
+    description?: string;
```
LeandroTorresSicilia, line 10: title and description must be a ReactNode, they can pass a string but they can also pass whatever they want with its own styles or a react intl FormattedMessage component for example

packages/listview/src/components/DownloadModal/index.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    /** A CSS class for the outer element, in addition to the component's base classes. */
+    className?: string;
+    /** An object with custom style applied to the outer element. */
+    style?: CSSProperties;
+    /** The name of the file to be downloaded. */
+    fileName?: string;
+    /** The maximum number of entries to be downloaded. */
+    maxEntries?: string | number;
```
LeandroTorresSicilia, line 18: reuse the Max defined above:
```
maxEntries?: Max;
```

packages/listview/src/components/DownloadModal/index.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    /** Controls whether the Modal is opened or not. If true, the modal is open. */
+    isOpen?: boolean;
+    /** The action triggered when the component request to close
+     *  (e.g click close button, press esc key or click outside the modal). */
+    onRequestClose?: () => void;
+    /** The action triggered when the component request to download.
+     * This is asynchronous function that must return the data to download in an array of objects format.
+     */
+    onDownload?: (max: Max, format: string) => Array<object>;
```
LeandroTorresSicilia, line 27: returning type here must be a promise:
```
... => Promise<Record<string, unknown>[]>
```
or
```
... => Promise<Array<Record<string, unknown>>>
```

packages/listview/src/components/DownloadModal/index.js
```diff
@@ -0,0 +1,120 @@
...
+};
+
+DownloadModal.defaultProps = {
+    title: 'Download',
+    description:
+        'The entries matching will be downloaded. If you need over 10,000 entries consider exporting.',
+    className: undefined,
+    style: undefined,
+    fileName: 'data',
+    maxEntries: '500',
```
LeandroTorresSicilia, line 114: default to number 500, instead of a string

PR #526, opened 2022-01-31
feat: implement column header filter text component

yvmunayev:


packages/listview/docs/stories/ColumnHeaderFilterText.story.js
```diff
@@ -0,0 +1,128 @@
...
+export const BasicColumnHeaderFilterText = () => {
+    return (
+        <Application>
+            <Container>
+                <Table data={dataTable} keyField="id" variant="listview">
+                    <Column header="Id" field="id" />
+                    <Column
+                        header="Name"
+                        field="name"
+                        sortable
```
LeandroTorresSicilia, line 46: remove sortable here and also let's make this example work

packages/listview/docs/stories/ColumnHeaderFilterText.story.js
```diff
@@ -0,0 +1,128 @@
...
+                >
+                    <Column header="Id" field="id" />
+                    <Column
+                        header="Name"
+                        field="name"
+                        sortable
+                        headerComponent={(props) => (
+                            <ColumnHeaderFilterText onFilter={setFilters} {...props} />
+                        )}
+                    />
```
LeandroTorresSicilia, line 111: can we use the component like this:
```
<Column
    header="Name"
    field="name"
    sortable
    headerComponent={ColumnHeaderFilterText}
    onFilter={setFilters}
/>
```
?

LeandroTorresSicilia:
@yvmunayev let's add an `index.d.ts` file add typing to this component

packages/listview/docs/stories/ColumnHeaderFilterText.story.js
```diff
@@ -0,0 +1,128 @@
...
+        if (filters.length > 0) {
+            return dataTable.filter((item) => {
+                return filters.some((word) => {
+                    const regex = new RegExp(word.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'i');
+                    return regex.test(item.name);
+                });
+            });
+        }
+        return dataTable;
+    }, [filters]);
```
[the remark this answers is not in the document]
yvmunayev: The thing about that helper is that it has support for multiple fields, but not for multiple queries.

I would have to do a mergeDedupe

packages/listview/docs/stories/ColumnHeaderFilterText.story.js
```diff
@@ -0,0 +1,122 @@
...
+                    <Column header="Id" field="id" />
+                    <Column header="Name" field="name" headerComponent={ColumnHeaderFilterText} />
+                    <Column header="Company" field="company" sortable />
+                    <Column header="Status" field="status" component={Status} />
+                    <Column header="Created At" field="createdAt" component={CreatedAt} />
+                </Table>
+            </Container>
+        </Application>
+    );
+};
```
LeandroTorresSicilia, line 51: here let's make the filter work

packages/listview/src/components/ColumnHeaderFilterText/FilterText/index.js
```diff
@@ -0,0 +1,91 @@
...
+    StyledButton,
+    StyledFooter,
+    StyledIconPlus,
+    StyledContent,
+} from './styled';
+import formatFilters from './helpers/formatFilters';
+
+function FilterText(props) {
+    const { onFilter, defaultFilters, headerText, onRequestClose } = props;
+    const [filters, setFilters] = useState(formatFilters(defaultFilters));
```
LeandroTorresSicilia, line 18: here this state is the same in the parent component, can we keep only one from the parent?

packages/listview/src/components/ColumnHeaderFilterText/FilterText/index.js
```diff
@@ -0,0 +1,91 @@
...
+                            <StyledInput
+                                placeholder={`Find ${headerText}`}
+                                onChange={(event) => handleChange(event, index)}
+                                value={value}
+                            />
+                            <RenderIf isTrue={filters.length > 1}>
+                                <ButtonIcon icon={<TrashFilled />} onClick={() => remove(index)} />
+                            </RenderIf>
+                        </StyledContentField>
+                    </>
```
LeandroTorresSicilia, line 52: here you need to use a wrapper and pass the key prop needed for the mapping

packages/listview/src/components/ColumnHeaderFilterText/index.js
```diff
@@ -0,0 +1,103 @@
...
+        onSort,
+        sortDirection,
+        sortable,
+        headerAlignment,
+    } = props;
+
+    const triggerRef = useRef();
+    const [isOpen, setIsOpen] = useState(false);
+    const [filters, setFilters] = useState(defaultFilters);
+    const [hasFilter, setHasFilter] = useState(defaultFilters.length > 0);
```
LeandroTorresSicilia, line 22: managing states introduce complexity, then we should try to avoid using states only when strictly necessary and stick to functional, here hasFilter can be computed from filters state

LeandroTorresSicilia:
let's add a clear all button that reset all filters, then chek with @TahimiLeonBravo how this will look

LeandroTorresSicilia:
when there is only one condition the remove button disappear, then I need to remove all in the input manually, I think this should not be hidden for the first input

PR #527, opened 2022-02-01
fix: download modal comments

amontalvof:


packages/listview/src/components/DownloadModal/helpers/convertToCSV.js
```diff
@@ -1,8 +1,12 @@
+import formatValue from './formatValue';
+
 const convertToCSV = (data) => {
     const array = [Object.keys(data[0])].concat(data);
     return array
         .map((item) => {
-            return Object.values(item).toString();
+            return Object.values(item)
+                .map((value) => formatValue(value))
```
LeandroTorresSicilia, line 8: here you can simply:
```
.map(formatValue)
```

packages/listview/src/components/DownloadModal/index.d.ts
```diff
@@ -1,21 +1,24 @@
-/* eslint-disable @typescript-eslint/ban-types */
-import { CSSProperties } from 'react';
+import { CSSProperties, ReactNode } from 'react';
 
 type Max = string | number;
 
+type ModalHeader = ReactNode | string;
```
LeandroTorresSicilia, line 5: you can use ReactNode directly since it includes the string type

LeandroTorresSicilia (approved):

PR #529, opened 2022-02-02
fix: improve AuditLogs component

HellWolf93:


packages/listview/src/components/AuditLogs/filterDropdown/filters.tsx
```diff
@@ -1,126 +1,102 @@
...
         const newValue = event.target.value;
-        const newLabels: Record<string, string[]> = { ...filters.labels };
-        const newValues = [...newLabels[nameInProps]];
-        newValues[index] = newValue;
-        newLabels[nameInProps] = newValues;
-        updateFilters({
-            ...filters,
-            labels: newLabels,
-        });
+        onChange(nameInProps, nameInProps, newValue);
```
LeandroTorresSicilia, line 45: here you are sending nameInProps twice

packages/listview/src/components/AuditLogs/filterDropdown/filters.tsx
```diff
@@ -1,126 +1,102 @@
...
     };
```
LeandroTorresSicilia, line 46: I see below you define this:
```onChange: (oldName: string, newName?: string, value?: string) => void;```
I think we should use an object thus we have better understanding of the prop names

LeandroTorresSicilia:
Screen Shot 2022-02-04 at 9 48 03 AM
picklist shows in this way when there are no more labels, we should show an empty message, talk to @TahimiLeonBravo about this

LeandroTorresSicilia:
when filter by `username: leo@nexxt` the filter does not found any match, however when filter by other it does it right also if I combine other with username it does it right but with username only it does not

LeandroTorresSicilia:
let's change date by createdAt

LeandroTorresSicilia:
Screen Shot 2022-02-04 at 9 51 39 AM
remove id from here

PR #532, opened 2022-02-05
fix: resolve auditlogs comments

HellWolf93:


packages/listview/src/components/AuditLogs/filterDropdown/filters.tsx
```diff
@@ -32,20 +33,29 @@ const LabelFilter = ({
...
+        });
 
     const handleValueChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         const newValue = event.target.value;
-        onChange(nameInProps, nameInProps, newValue);
+        onChange({
+            oldName: nameInProps,
+            newName: nameInProps,
+            value: newValue,
+        });
```
HellWolf93, line 54: @LeandroTorresSicilia this is intentional. The onChange function updates the input value when the `oldName` and `newName` are equal.

PR #542, opened 2022-03-03
fix: improved country picker

amontalvof:


packages/forms/docs/stories/CountryPicker.story.js
```diff
@@ -18,6 +18,22 @@ export const Basic = () => {
     );
 };
 
+export const FilteredCountries = () => {
+    const [value, setValue] = useState({ isoCode: 'mx' });
+    const countries = ['cu', 'mx', 'es', 'us'];
```
LeandroTorresSicilia, line 23: we can put this outside thus avoiding rerenders

packages/forms/src/components/CountryPicker/index.js
```diff
@@ -120,6 +104,250 @@ CountryPicker.defaultProps = {
     readOnly: false,
     required: false,
     error: undefined,
+    countries: [
```
LeandroTorresSicilia, line 107: I think by default we should use undefined and internally when it is not an array then we always return all the countries

PR #577, opened 2022-07-23
fix: add parameters types to mutate on useMutationFlow

### LINKED ISSUE #572: FIx useMutationFlow types
The returned mutate function is defined as with no argument, then sometimes you need to pass an argument for example:

```
const useRejectRideOffer = () => {
    const { fetchAsync, isLoading } = useAuthFetch({
        baseUrl: BASE_URL,
    });
    const { mutate } = useMutationFlow({
        mutation: async (rideId: string): Promise<Result> => {
            const result = await fetchAsync<Result>(`/ride-requests/${rideId}/reject`, {
                method: 'PUT',
            });
            return result;
        },
        successMessage: 'Ride rejected sucessfully.',
        errorMessage: 'There was an error.Please try again.',
    });

    return useMemo(() => ({
        rejectRide: async (rideId: string) => {
            const result = await confirmModal({
                variant: 'destructive',
                header: 'Reject Ride',
                question: "This ride will be rejected. You can't undo this action.",
                okButtonLabel: 'Reject',
            });
            if (result) {
                // here we get a type error because the mutate func returned by useMutationFlow does not expect any argument
                await mutate(rideId);
            }
        },
        isLoading,
    }), [isLoading, mutate]);
};
```

HellWolf93:
fix #572

### DIFF
--- a/packages/hooks/src/useMutationFlow/index.d.ts
+++ b/packages/hooks/src/useMutationFlow/index.d.ts
@@ -17,4 +17,4 @@ export interface MutationOptions {
     type?: feedbackType | Type;
 }
 
-export default function (opts: MutationOptions): { mutate: () => void };
+export default function (opts: MutationOptions): { mutate: (...args: any[]) => void };
--- a/packages/hooks/src/useMutationFlow/index.d.ts
+++ b/packages/hooks/src/useMutationFlow/index.d.ts
@@ -17,4 +17,4 @@ export interface MutationOptions {
     type?: feedbackType | Type;
 }
 
-export default function (opts: MutationOptions): { mutate: (...args: any[]) => void };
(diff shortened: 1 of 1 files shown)


packages/hooks/src/useMutationFlow/index.d.ts
```diff
@@ -17,4 +17,4 @@ export interface MutationOptions {
     type?: feedbackType | Type;
 }
 
-export default function (opts: MutationOptions): { mutate: () => void };
+export default function (opts: MutationOptions): { mutate: (...args: any[]) => void };
```
LeandroTorresSicilia, line 20: instead of any we should use unknown

LeandroTorresSicilia (approved):

PR #581, opened 2022-07-28
fix: filters alignment

amontalvof:


packages/listview/src/components/AuditLogs/header.tsx
```diff
@@ -112,9 +112,10 @@ const ToolBar = () => {
 
 const AuditLogsHeader = ({ title }: AuditLogsHeaderProps): JSX.Element => {
     const isTitleString = typeof title === 'string';
+    const notHaveTitle = !title;
```
LeandroTorresSicilia, line 115: negative sentence are more difficult to read then lets use `hasTitle` instead

PR #596, opened 2022-08-25
fix: improve types of firebase hooks

HellWolf93:


packages/firebase-hooks/src/firestore/types.d.ts
```diff
@@ -0,0 +1,13 @@
+export type Doc<DocType> = {
+    [key in keyof DocType]: DocType[key];
+} & {
+    id: number;
```
LeandroTorresSicilia, line 4: if this is the id from firestore then it is a string

LeandroTorresSicilia (approved):

PR #605, opened 2022-09-23
fix: button link in rainbow-modules does not work with theme

yvmunayev:
https://github.com/nexxtway/rainbow-modules/issues/604


packages/misc/src/components/LinkButton/styled.js
```diff
@@ -24,7 +24,7 @@ export const StyledLink = attachThemeAttrs(
         );
     }),
 ).attrs((props) => {
-    if (props.palette.isDark) {
+    if (props.theme.rainbow.palette.isDark) {
```
LeandroTorresSicilia, line 27: here we can safely use `props.palette.isDark` because of `attachThemeAttrs`

Issue #621, opened 2022-10-29
feat: add cell currency component using intl formatter


yvmunayev:
feat: add cell currency component using intl formatter

PR #622, opened 2022-10-29
feat: add percent column component using intl formatter

yvmunayev:
#621 

feat: add percent column component using intl formatter


LeandroTorresSicilia:
@yvmunayev here we need to support cellAlignment prop defined in Column component thus we can align to left, center or right
Also percents by default must be aligned to right

packages/listview/src/components/PercentColumn/index.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import useLocale from 'react-rainbow-components/libs/hooks/useLocale';
```
LeandroTorresSicilia, line 3: we can't do this anymore, I have refactored all this imports because this causes issues with Next JS and SSR
In this case now you can use ```import { useLocale } from 'react-rainbow-components';```

packages/listview/src/components/PercentColumn/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const PercentColumn: React.FC<PercentColumnProps> = (props: PercentColumnProps) => {
+    const { value, locale: localeProp, className, style, ...rest } = props;
+    const locale = useLocale(localeProp);
+    const content = formatPercent(value ?? 0, locale, rest);
+
+    return (
+        <StyledCellContainer className={className} style={style}>
+            <span title={content}>{content}</span>
+        </StyledCellContainer>
```
LeandroTorresSicilia, line 16: I think maybe here we can get rid off the span and put the title directly into the StyledCellContainer

PR #623, opened 2022-10-29
feat: add currency column component using intl formatter

yvmunayev:
#621 

feat: add currency column component using intl formatter


LeandroTorresSicilia:
@yvmunayev same comments as in the percent one

Issue #625, opened 2022-10-30
fix: show app message params type


yvmunayev:
fix: show app message params type

LeandroTorresSicilia:
closed by #629

PR #629, opened 2022-10-30
fix: show app message type

yvmunayev:
#625 

fix: show app message type


packages/app/src/components/AppMessage/index.js
```diff
@@ -32,7 +32,7 @@ AppMessage.defaultProps = {
     isVisible: false,
     message: undefined,
     variant: 'error',
-    timeout: undefined,
+    timeout: 5000,
```
LeandroTorresSicilia, line 35: this change is a breaking change, so we need to discuss it

Issue #618, opened 2022-10-18
feat: add openUpdateUserPhotoModal method in storage module


LeandroTorresSicilia:
openUpdateUserPhotoModal will be a method that you can call anywhere in the app and it will show a modal, for this let's use a portal that is created at the moment of calling the method

```
openUpdateUserPhotoModal({
   bucket: string, // optional, by default user the default bucket
   path:  string, // optional, default `${auth.tenantId}/users/${uid}/profilePhoto/${nanoid()}`
   onPhotoUpdated: () => {}, // optional
});
```

for example if you use ```path={`users/${uid}`}```
then internally we do use `${path}/${nanoid()}`

this is for concatenating the path the user pass when using the method with the  nanoid that will generate the file name to avoid collision in filenames

https://www.figma.com/file/bMnIxYEbFqybEemecEYHLt/rainbow-modules?node-id=1%3A563

PR #643, opened 2022-11-08
feat: implement openUpdateUserPhotoModal

LeandroTorresSicilia:
fix: #618


packages/storage/docs/components/UpdateUserPhotoModal.story.mdx
```diff
@@ -0,0 +1,19 @@
+import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
+import UpdateUserPhotoModal from '../../src/components/UpdateUserPhotoModal';
+import * as stories from '../stories/UpdateUserPhotoModal.story.js';
+
+<Meta title="Modules/Storage/Components" component={UpdateUserPhotoModal} />
+
+# Overview
+
+`UpdateUserPhotoModal` is a component to manage and select images stored in Firebase Storage.
```
LeandroTorresSicilia, line 9: improve description here

packages/storage/docs/stories/UpdateUserPhotoModal.story.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { RainbowFirebaseApp } from '@rainbow-modules/app';
+import UpdateUserPhotoModal from '../../src/components/UpdateUserPhotoModal';
+import app from '../../../../firebase';
+
+export const basicUpdateUserPhotoModal = () => {
+    const defaultSrc = 'https://www.w3schools.com/howto/img_avatar.png';
+    return (
+        <RainbowFirebaseApp app={app}>
+            <UpdateUserPhotoModal path={defaultSrc} />
```
LeandroTorresSicilia, line 10: improve example check with @TahimiLeonBravo

packages/storage/package.json
```diff
@@ -8,7 +8,9 @@
     ],
     "dependencies": {
         "@rainbow-modules/firebase-hooks": "^0.55.0",
-        "@rainbow-modules/icons": "^0.55.0"
+        "@rainbow-modules/icons": "^0.55.0",
+        "react-cropper": "^2.1.8",
+        "react-rainbow-components": "1.31.0-canary.6ccebb7"
```
LeandroTorresSicilia, line 13: remove this dependency here

packages/storage/src/components/UpdateUserPhotoModal/styled.js
```diff
@@ -0,0 +1,158 @@
...
+    width: none;
+    padding: 0;
+
+    [data-id*='modal-content'] {
+        padding: 0;
+    }
+
+    .byCVyc {
+        padding: 0;
+    }
```
LeandroTorresSicilia, line 85: remove this

packages/storage/src/components/UpdateUserPhotoModal/styled.js
```diff
@@ -0,0 +1,158 @@
...
+        padding-right: 90px;
+    }
+`;
+
+export const StyledParagraph = styled.p`
+    font-weight: 600;
+    font-size: 14px;
+    line-height: 150%;
+    margin-bottom: 35px;
+    color: #4f5657e5;
```
LeandroTorresSicilia, line 36: let's no use colors inline

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,135 @@
...
+                        assistiveText="Jose Leandro"
+                        title="Jose Leandro"
+                        size="large"
+                    />
+                    <Button
+                        id="button-1"
+                        variant="neutral"
+                        label="Change Profile Picture"
+                        onClick={handleOnClick}
+                    />
```
LeandroTorresSicilia, line 66: remove this, leave only the modal

LeandroTorresSicilia:
This case does not look like in design:
Screen Shot 2022-12-02 at 2 25 08 PM

LeandroTorresSicilia:
When there is no displayName instead of use initials let's use the user icon from rainbow-modules/icons

packages/storage/docs/components/UpdateUserPhotoModal.story.mdx
LeandroTorresSicilia: @adrianestevez1 this is not solved

packages/storage/src/components/UpdateUserPhotoModal/styled.js
LeandroTorresSicilia: @adrianestevez1 this color here is not valid and also we can't use colors inline

packages/storage/src/components/UpdateUserPhotoModal/editRemoveDialog.js
```diff
@@ -0,0 +1,101 @@
...
+            <StyledParagraph>
+                A picture helps people recognize you and lets you know when you&apos;re sign in to
+                your account.
+            </StyledParagraph>
+            <StyledAvatar src={imageSrc} initials={avatarInitials} />
+            <input
+                type="file"
+                ref={hiddenFileInput}
+                onChange={onFileChange}
+                style={{ display: 'none' }}
```
LeandroTorresSicilia, line 58: no inline styles

LeandroTorresSicilia:
Here you forgot the go back button:

Screen Shot 2022-12-02 at 2 41 53 PM

packages/storage/docs/stories/UpdateUserPhotoModal.story.js
```diff
@@ -0,0 +1,100 @@
...
+                        />
+                        <StyledButton id="button-1" variant="neutral" onClick={handleOnClick}>
+                            <Camera />
+                        </StyledButton>
+                    </div>
+                </div>
+                <UpdateUserPhotoModal
+                    photo={photoURL}
+                    bucket="video-labeling"
+                    path="usertest"
```
LeandroTorresSicilia, line 81: here let's use the default values, so remove path and bucket properties

LeandroTorresSicilia:
@TahimiLeonBravo  When remove a photo what we will do?

packages/storage/src/index.d.ts
```diff
@@ -1,2 +1,3 @@
 // eslint-disable-next-line import/prefer-default-export
```
LeandroTorresSicilia, line 1: now you can remove this eslint exception

packages/storage/docs/stories/UpdateUserPhotoModal.story.js
```diff
@@ -0,0 +1,100 @@
...
+                            title="Jose Leandro"
+                            size="large"
+                        />
+                        <StyledButton id="button-1" variant="neutral" onClick={handleOnClick}>
+                            <Camera />
+                        </StyledButton>
+                    </div>
+                </div>
+                <UpdateUserPhotoModal
+                    photo={photoURL}
```
LeandroTorresSicilia, line 79: this can be solved internally then we don't need to have this prop in the public API, remember always that this public API is the most important thing in component design, a future change in the API can make Breaking Changes which we need to avoid, then any prop that need to change, add or remove we need to discuss it

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+                    </StyledCropperContainer>
+                </StyledModalEdit>
+            </RenderIf>
+        </>
+    );
+}
+
+UpdateUserPhotoModal.propTypes = {
+    path: PropTypes.string,
+    photo: PropTypes.string,
```
LeandroTorresSicilia, line 178: same ^^

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+    bucket: PropTypes.string,
+    isOpen: PropTypes.bool,
+    onRequestClose: PropTypes.func,
+    avatarInitials: PropTypes.string,
```
LeandroTorresSicilia, line 182: same ^^ this can be used inside the component, not need for this prop

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+    handlePhotoUrl: PropTypes.func,
```
LeandroTorresSicilia, line 183: instead of `handlePhotoUrl` we will use `onPhotoUpdated`

LeandroTorresSicilia:
@adrianestevez1 add the index.d.ts with the types definitions for the component

packages/storage/docs/stories/UpdateUserPhotoModal.story.js
```diff
@@ -0,0 +1,100 @@
...
+
+const StyledAvatar = styled(Avatar)`
+    width: 70px;
+    height: 70px;
+`;
+
+export const BasicUpdateUserPhotoModal = () => {
+    const [photoURL, setPhotoURL] = useState();
+    const [isOpen, setIsOpen] = useState(false);
+    const [initials, setInitials] = useState();
```
LeandroTorresSicilia, line 35: Instead of using this states (photoURL and initials) you can use the useCurrentUser hook which will return null if unauthenticated or the user object when it is authenticated, then you can get the phototURL directly from the user object and the initials can be computed inline, thus avoiding using states.
Keeping track of states it is hard and lead to bugs, then always you need to think it twice to decide to use a state, in this case the isOpen state is needed but the others two do not.

packages/storage/docs/stories/UpdateUserPhotoModal.story.js
```diff
@@ -0,0 +1,100 @@
...
+
+    const getInitials = (name) => {
+        if (name) {
+            const names = name.split(' ');
+            const initials =
+                names[0].substring(0, 1).toUpperCase() + names[1].substring(0, 1).toUpperCase();
+            return initials;
+        }
+        return '';
+    };
```
LeandroTorresSicilia, line 45: since this function does not depend on any prop or state it can be declared outside

packages/storage/src/index.js
```diff
@@ -1,2 +1,3 @@
 // eslint-disable-next-line import/prefer-default-export
```
LeandroTorresSicilia, line 1: same here remove this eslint exception

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+export default function UpdateUserPhotoModal(props) {
+    const { path, photo, bucket, isOpen, onRequestClose, avatarInitials, handlePhotoUrl } = props;
+
+    const app = useFirebaseApp();
+    const { user, reload } = useCurrentUserState() || {};
+
+    const pathProp = path || `users/${user ? user.uid : ''}/profilePhoto`;
+    const bucketProp = bucket || `${app.options.storageBucket}`;
+
+    const [photoURL, setPhotoURL] = useState(photo);
```
LeandroTorresSicilia, line 36: same here I think this prop is not needed

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+        onRequestClose();
+        setIsOpenEdit(true);
+    };
+
+    const onCrop = async () => {
+        const imageElement = cropperRef ? cropperRef.current : null;
+        const cropper = imageElement ? imageElement.cropper : null;
+        const storage = getStorage(app, bucketProp);
+        const photoRef = ref(storage, `${pathProp}/${nanoid()}`);
+        const blob = await fetch(cropper.getCroppedCanvas().toDataURL()).then((r) => r.blob());
```
LeandroTorresSicilia, line 51: if you use async await then you don't need to use `then`

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+        uploadTask.on(
+            'state_changed',
+            (snapshot) => {
+                const progress = (snapshot.bytesTransferred / snapshot.totalBytes) * 100;
+                if (progress !== 100) {
+                    setLoading(true);
+                }
+            },
+            (error) => {
+                console.log(error);
```
LeandroTorresSicilia, line 63: check with @TahimiLeonBravo how handle errors

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+        const imageElement = cropperRef ? cropperRef.current : null;
+        const cropper = imageElement ? imageElement.cropper : null;
+        cropper.rotate(90);
+    };
+
+    const handleOnClickRemove = async () => {
+        const storage = getStorage(app);
+        const photoRef = ref(storage, photoURL);
+        await deleteObject(photoRef)
+            .then(() => {
```
LeandroTorresSicilia, line 89: same ^^ don't use .then and .catch and instead do:
```
try {
   await deleteObject(photoRef)
   ...
} catch (error) {
   ...
}
```

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+                updateProfile(user, {
+                    photoURL: '',
+                }).then(() => {
+                    setPhotoURL('');
+                    setIsOpenEdit(false);
+                    onRequestClose();
+                    reload().then(() => {
+                        handlePhotoUrl('');
```
LeandroTorresSicilia, line 97: for now send null here when remove the photo

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+                    });
+                });
+            })
+            .catch((error) => {
+                console.log(error);
```
LeandroTorresSicilia, line 102: same talk to @TahimiLeonBravo to handle errors

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+
+    return (
+        <>
+            <StyledModal isOpen={isOpen} onRequestClose={onRequestClose}>
+                <EditRemoveDialog
+                    imageSrc={photo}
+                    avatarInitials={avatarInitials}
+                    handleOnClickEdit={handleOnClickEdit}
+                    handleImageSrc={setPhotoURL}
+                    handleOnClickRemove={handleOnClickRemove}
```
LeandroTorresSicilia, line 114: normally for convention when you can to pass events to props you do this:
```
onClick={handleOnClick}
```
the word handle is used for  the handler but for the prop use onSomethign attaching the on always, like with native events

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,194 @@
...
+    const app = useFirebaseApp();
+    const { user, reload } = useCurrentUserState() || {};
+
+    const pathProp = path || `users/${user ? user.uid : ''}/profilePhoto`;
+    const bucketProp = bucket || `${app.options.storageBucket}`;
+
+    const [photoURL, setPhotoURL] = useState(photo);
+    const [isOpenEdit, setIsOpenEdit] = useState(false);
+    const cropperRef = useRef(null);
+    const [loading, setLoading] = useState(false);
```
LeandroTorresSicilia, line 39: normally for booleans we have a convention to use the `is` word as prefix in this case you should use `isLoading` this is for indicate that it is a boolean

packages/storage/src/components/UpdateUserPhotoModal/editRemoveDialog.js
```diff
@@ -0,0 +1,101 @@
...
+        handleImageSrc,
+        avatarInitials,
+    } = props;
+    const hiddenFileInput = React.useRef(null);
+
+    const handleEdit = () => {
+        hiddenFileInput.current.click();
+    };
+
+    function readFile(file) {
```
LeandroTorresSicilia, line 29: here to be consisten use const arrow function

TahimiLeonBravo:
> 
@LeandroTorresSicilia @adrianestevez1 
This is the design for the remove photo flow
https://www.figma.com/proto/bMnIxYEbFqybEemecEYHLt/rainbow-modules?page-id=0%3A1&node-id=1%3A563&viewport=515%2C522%2C0.17&scaling=scale-down&starting-point-node-id=1%3A563&show-proto-sidebar=1

packages/storage/src/components/UpdateUserPhotoModal/index.js
TahimiLeonBravo: @LeandroTorresSicilia @adrianestevez1 
This is the design for the errors
https://www.figma.com/file/bMnIxYEbFqybEemecEYHLt/rainbow-modules?node-id=150%3A1036&t=yDG1pijxXNgFmCTy-1

packages/storage/docs/components/UpdateUserPhotoModal.story.mdx
TahimiLeonBravo: @LeandroTorresSicilia @adrianestevez1 
Let's use this description ;) 
"UpdateUserPhotoModal is a component for managing the user profile photo. It is built using Firebase, and all images will be stored on Firebase Storage."

packages/storage/src/components/UpdateUserPhotoModal/index.d.ts
```diff
@@ -0,0 +1,18 @@
+export interface ImageGalleryProps {
```
LeandroTorresSicilia: ImageGalleryProps?

packages/storage/src/components/UpdateUserPhotoModal/index.d.ts
```diff
@@ -0,0 +1,15 @@
...
+    path: string;
+    /** The bucket where the images are going to be stored */
+    bucket: string;
+    /** Controls whether the Modal is opened or not. If true, the modal is open. */
+    isOpen: boolean;
+    /** The action triggered when the component request to close
+     *  (e.g click close button, press esc key or click outside the modal). */
+    onRequestClose: () => void;
+    /** The action triggered when the user click the save button */
+    onPhotoUpdated: (photo: string) => void;
```
LeandroTorresSicilia, line 12: all of these props must be optionals

packages/storage/src/components/UpdateUserPhotoModal/index.js
```diff
@@ -0,0 +1,233 @@
...
+            try {
+                await deleteObject(photoRef);
+                await updateProfile(user, {
+                    photoURL: '',
+                });
+                setIsOpenEdit(false);
+                await reload();
+                onPhotoUpdated(null);
+            } catch (error) {
+                showAppMessage({
```
LeandroTorresSicilia, line 126: look at the design that tahimi attached, it is not an app message but a custom error message inside the modal

Issue #670, opened 2023-01-01
feat: implement `borderRadius` on `UniversalFormModal` component


TahimiLeonBravo:

TahimiLeonBravo:
### Requirements:
- It should change the `border-radius` on the footer buttons
- It should change the `border-radius` on the modal

PR #671, opened 2023-01-05
feat: implement borderRadius on UniversalFormModal component

yvmunayev:
fix: #670

### DIFF
--- a/packages/forms/docs/stories/UniversalFormModal.story.js
+++ b/packages/forms/docs/stories/UniversalFormModal.story.js
@@ -211,7 +211,7 @@ const App = () => {
                 />
             </TitleContainer>
             <Books />
-            <UniversalFormModal fields={Fields} {...connectedModalProps} />
+            <UniversalFormModal fields={Fields} {...connectedModalProps} borderRadius="square" />
         </div>
     );
 };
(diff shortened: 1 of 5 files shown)


packages/forms/docs/stories/UniversalFormModal.story.js
```diff
@@ -211,7 +211,7 @@ const App = () => {
                 />
             </TitleContainer>
             <Books />
-            <UniversalFormModal fields={Fields} {...connectedModalProps} />
+            <UniversalFormModal fields={Fields} {...connectedModalProps} borderRadius="square" />
```
LeandroTorresSicilia, line 214: here the old example is changed, let's revert it to use the default border

packages/forms/src/components/UniversalFormModal/index.d.ts
```diff
@@ -28,6 +28,8 @@ export interface UniversalFormModalProps {
     cancelButtonLabel?: ReactNode;
     /** The id of the outer element. */
     id?: string;
+    /** The border radius of the button and dropdown. Valid values are square, 'semi-square', semi-rounded and rounded. This value defaults to rounded. */
```
LeandroTorresSicilia, line 31: button and dropdown??

LeandroTorresSicilia (approved):

PR #680, opened 2023-02-17
feat: multi phone number component first commit

maxxgreene:


packages/forms/src/components/MultiPhoneNumberInput/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+        const isFirst = index === 0;
+        const inputConfig = onAddPhoneNumber
+            ? onAddPhoneNumber(index)
+            : [{ label: isFirst ? 'Primary' : 'Secondary' }, { label: 'Note' }];
+
+        const rowError = !error || typeof error === 'string' ? null : (error[index] as RowError);
+        const phoneError = rowError && rowError.phone;
+        const noteError = rowError && rowError.note;
+        return (
+            <Fieldset>
```
LeandroTorresSicilia, line 63: here we need to pass a key

packages/forms/src/components/MultiPhoneNumberInput/utils.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable import/prefer-default-export */
+let counter = 0;
+export const uniqueId = (): string => {
```
LeandroTorresSicilia, line 3: I don't see where is this being used

